PR #21040, opened 2026-04-18
LeetCode hard series: start with Sudoku Solver across the full matrix

tamnd:
This starts a dedicated LeetCode PR after landing the SPOJ work.

I wanted the first batch here to be a real hard problem instead of branch cleanup, so this PR opens with `0037-sudoku-solver` across the full language matrix under `tests/leetcode/x/*`.

What is in this batch:
- a short `0037.md` write-up for every target language folder
- `0037.in` and `0037.out` fixtures
- implementations for C, C++, C#, Clojure, Dart, Elixir, Erlang, Go, Haskell, Java, Kotlin, Lean, Lua, Mochi, OCaml, Pascal, PHP, Python, Ruby, Rust, Scala, Swift, TypeScript, and Zig
- `TASKS.md` updated so the hard-problem sequence is tracked from here

Problem notes:
- Sudoku Solver is a plain backtracking problem, but it is a good stress test for the runners because every language needs the same board mutation and validity checks to behave the same way.
- I used the standard search order: find the next empty cell, try digits `1..9`, reject by row/column/box constraints, and backtrack when a branch fails.

Validation:
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0037 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0037$'`

I’ll keep extending this PR one problem at a time in index order from the hard section.


tamnd:
Started this PR with `0037-sudoku-solver` because it feels like a good hard reset after the SPOJ branch.

What I like about this problem is that the idea is simple to explain but annoying to get wrong in twenty-plus runtimes. The core loop is just: scan for the next `.` cell, try digits `1` through `9`, reject anything that already appears in the same row, column, or 3x3 box, and backtrack when a choice paints us into a corner.

The interesting part of this batch was not the algorithm itself, it was making the same board mutation story behave consistently everywhere. A few runtimes needed cleanup around recursion, parser quirks, or string/character handling before they all agreed on the same solved board.

This one is fully validated with the isolated `0037` slow test run across the whole language matrix, including Lean and Scala. Next I’m moving to `0041-first-missing-positive`.

tamnd:
Added `0041-first-missing-positive`.

This is one of those hard problems where the trick is really about trusting the array as your bookkeeping structure. The useful observation is that for an array of length `n`, the only values that matter are `1..n`. So instead of tracking presence in a separate set, I keep swapping each value `v` into the slot where it belongs, index `v - 1`, until every reachable value is either in the right place or blocked by a duplicate / out-of-range number.

After that cleanup pass, the answer almost reads itself: scan from the front, and the first index where `nums[i] != i + 1` is the missing positive. If every slot lines up, then the missing value is `n + 1`.

This batch was mostly about getting the same in-place mutation story to behave across all runtimes. The algorithm is short, but low-level details like tuple updates, stdin handling, and compiler-specific newline parsing still needed a round of cleanup before the whole matrix agreed.

Validated with:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0041 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0041$'`

Next hard problem is `0042-trapping-rain-water`.

tamnd:
Added `0042-trapping-rain-water`.

I went with the two-pointer solution here because it keeps the core insight visible without building extra prefix/suffix arrays. The idea is that the side with the smaller current wall is the only side we can settle safely at that moment. So we keep `leftMax` and `rightMax`, advance the side whose bound is smaller, and either add trapped water or raise that side’s max.

This batch also had a good reminder that fixture quality matters as much as code quality. The first isolated run looked like a bunch of runtime bugs, but the real source was a malformed last test case in the shared input. Once that was fixed, most of the failures disappeared and the remaining work was just the usual language-specific cleanup around newline handling and Lean recursion.

Validated with:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0042 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0042$'`

Next hard problem is `0044-wildcard-matching`.

tamnd:
Added `0044-wildcard-matching`.

I used the classic greedy wildcard matcher here instead of the full DP table. The useful mental model is that only `*` creates real ambiguity, so rather than remembering every subproblem, we only remember the most recent star position and the string index where we started letting that star expand. On a mismatch, if we have seen a star before, we back up to just after that star and let it absorb one more character. That keeps the implementation compact and avoids the giant `O(nm)` table.

This batch was mostly about empty-string and parser edge cases. The algorithm itself is short, but `0044` is a good stress test for runtimes that don’t love optional line consumption, embedded wildcard characters, or generated newline literals. The full pass is green now across the whole matrix.

Validated with:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0044 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0044$'`

Next hard problem is `0051-n-queens`.

tamnd:
Worked through `0051-n-queens` next.

This one is the classic backtracking shape, but the interesting part is making the search cheap enough that the same idea stays readable in every language. I went with the usual row-by-row placement and three occupancy trackers: one for columns, one for `row + col`, and one for the other diagonal. That way each candidate queen is just a few boolean checks instead of scanning the whole board.

The fixture prints the full set of boards, so this batch was a little more annoying than the count-only problems. A few runtimes were fine on the search itself but got tripped up by output order, separator handling, or language-specific string / collection quirks. The slowest cleanup was making sure Lean, Mochi, OCaml, and Zig all produced the exact same board listing format.

Validation passed on the full matrix for `0051`, so I’m moving on to `0052-n-queens-ii` next.

tamnd:
Picked up `0052-n-queens-ii` next.

This is really the counting version of the previous problem, so I kept the same backtracking shape: place one queen per row, track used columns and diagonals, and recurse. The only real algorithm change is that instead of collecting boards, each recursive branch returns a count and the caller sums them. That makes the code smaller and avoids carrying full board state into the output path.

Most of the work in this batch was not the search itself. Since `0051` had a custom board-printing fixture, I reused that code as the starting point and then stripped every runtime back down to plain counts. The first validation pass flushed out a bunch of leftover `=` separators and one broken C rewrite, then the second pass narrowed it down to a handful of runtimes with one-line output quirks. After that cleanup, the full matrix passed.

`0052` is green now, and I’m moving on to `0060-permutation-sequence` next.

tamnd:
Worked through `0060-permutation-sequence` next.

This one is a nice change of pace after the N-Queens pair. The key observation is that you do not need to generate permutations at all. Once you know there are `(n - 1)!` permutations starting with each possible first digit, you can treat `k` as an index into fixed-size blocks: pick the right digit for the current position, remove it, reduce `k` inside that block, and repeat.

So the implementation in every language follows the same shape: build the digit pool `1..n`, precompute factorials, convert `k` to zero-based, and then greedily remove one digit at a time. The main cleanup in this batch was not algorithmic. The first full pass came back with only one failure, in Mochi, where the local string/list construction style produced `nil` values. After switching that version to use substring-based digit extraction like the other working Mochi fixtures, the whole matrix passed.

`0060` is green now, and I’m moving on to `0065-valid-number` next.

tamnd:
Worked through `0065-valid-number` next.

This one looks annoying at first because the definition is written in a grammar-heavy way, but the implementation is simpler if you treat it as a one-pass state check. I used the same small set of flags in every language: have we seen any digit yet, have we seen a dot, have we seen an exponent, and if we did see an exponent, have we seen a digit after it. Then each character just updates or rejects based on those rules.

Most of the effort in this batch was translation cleanup rather than algorithm design. The first full pass narrowed it down to a handful of runtimes with local issues: Elixir and Erlang both collided with built-in numeric predicates by name, Kotlin and Scala had compressed-expression syntax damage from generation, Lean’s loop shape was wrong, and Zig needed a stdlib trim adjustment. Once those were rewritten more plainly, the full matrix passed cleanly.

`0065` is green now, and I’m moving on to `0068-text-justification` next.

tamnd:
Worked through `0068-text-justification` next.

This one is mostly about getting the formatting contract exactly right. The greedy part is simple: keep adding words to the current line while they still fit. The real work is in how you turn one chosen group of words into a line. If it is the last line, or if there is only one word, you join with single spaces and pad on the right. Otherwise you compute how many spaces are left to distribute, divide them across the gaps, and give the earlier gaps one extra space when it does not divide evenly.

For the fixture, I wrapped each output line in `|...|` and printed the line count per testcase. That made the trailing spaces visible and testable across all languages, which matters a lot for this problem. The first matrix run only exposed three runtime-specific issues: Clojure had malformed generated syntax, Lean needed the recursive helpers marked more explicitly, and Zig needed a current stdlib trim path plus one runtime integer fix. After those cleanups, the full matrix passed.

`0068` is green now, and I’m moving on to `0076-minimum-window-substring` next.

tamnd:
Finished `0076 Minimum Window Substring`.

This one is the classic moment where the problem looks like it wants brute force, but the real trick is to stop thinking in terms of "try every substring" and instead keep one window moving across the string. I track how many characters from `t` are still missing, expand the right side until the window becomes valid, and then immediately start shrinking from the left while it stays valid. The answer only changes during that shrinking phase.

What I like about this problem is that the important insight is not the counting map by itself. It is the discipline of knowing when the window is allowed to grow and when it must shrink. Once that is right, the code in each language becomes mostly bookkeeping.

I ran the isolated `0076` batch across the full language set before pushing this one.

tamnd:
Finished `0084 Largest Rectangle in Histogram`.

This one is a good reminder that the hard part is not the multiplication, it is deciding what rectangle each bar is actually allowed to support. For this batch I kept the implementation deliberately direct: pick a left edge, stretch the right edge outward, keep the smallest height seen so far, and compute the area for that whole span. The running minimum is what turns the interval into a valid rectangle.

What makes the problem interesting is that every new bar can only keep or lower that minimum, so you are constantly balancing width against height. Even with the simple scan, that tradeoff is the whole story of the problem.

I ran the isolated `0084` batch across the full language matrix before pushing it.

tamnd:
Finished `0085 Maximal Rectangle`.

This one is really `0084` wearing a matrix disguise. The useful shift is to stop trying to search rectangles directly in two dimensions. Instead, walk row by row, build the running column heights of consecutive `1`s, and treat each row as a histogram. Once you do that, every row asks the same question: what is the best rectangle ending here?

What I like about this problem is that the matrix version looks much more complicated than it really is. The main work is choosing the right representation after each row update. After that, the rectangle logic becomes familiar.

I ran the isolated `0085` batch across the full language matrix before pushing it.

tamnd:
Finished `0087 Scramble String`.

This one is one of those problems where the definition feels abstract until you reduce it to a very concrete question: if I split both strings at some position, can the left and right halves match either in the same order or swapped order? From there the whole problem becomes recursion on smaller intervals.

The important guardrail is the letter-count check before recursing. Without that, the search tree gets noisy fast. With it, most impossible branches die immediately and the recursive structure becomes much easier to trust.

I ran the isolated `0087` batch across the full language matrix before pushing it.

tamnd:
Finished `0097 Interleaving String`.

This one is a clean dynamic programming problem once you stop thinking about “alternating chunks” and just ask a smaller prefix question: can the first `i + j` characters of `s3` be built from the first `i` of `s1` and the first `j` of `s2`? That turns the whole problem into a table where each cell is only looking one step up or one step left.

What I like here is that the recurrence is very honest. If the next character of `s3` matches the next character from `s1`, you can come from above. If it matches the next character from `s2`, you can come from the left. If neither path works, the cell stays false.

I ran the isolated `0097` batch across the full language matrix before pushing it.

PR #21041, opened 2026-04-18
Continue LeetCode batches with Triangle across the language matrix

tamnd:
## Summary
This PR starts the next LeetCode batch after the previous series landed.

The first problem here is `0120 Triangle`, implemented across the full language matrix under `tests/leetcode/x/*`, with matching `.md`, `.in`, and `.out` files.

## What changed
- add `0120-triangle` solutions for all supported LeetCode test languages
- add the problem note explaining the bottom-up DP approach
- mark `0120-triangle` complete in `tests/leetcode/TASKS.md`

## Validation
Ran the isolated full-language check for `0120`:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0120 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0120$'
```

This passed locally.

## Notes
I will keep extending this branch one problem at a time from here.


tamnd:
Started the next LeetCode branch with `0120 Triangle`.

This one is a clean bottom-up DP problem. The useful observation is that once you stand on some cell in row `r`, the only future choices are the two cells directly below it in row `r + 1`. So instead of exploring paths from the top and carrying a lot of state around, I treated the last row as the initial DP and collapsed the triangle upward one row at a time.

For each position, the transition is just:
`triangle[r][c] + min(dp[c], dp[c + 1])`

That keeps the implementation small in every language and avoids recursion-heavy edge cases. The fixture also includes a negative-value case and a single-row case, since both are easy places for generated ports to drift.

Local validation passed for the full isolated language matrix for `0120`, including the slower Lean and Scala paths.

tamnd:
Added `0121 Best Time to Buy and Sell Stock`.

This one looks like it wants some kind of DP table at first, but the problem gets much simpler once you phrase it the right way: if today is the selling day, the only thing that matters from the past is the cheapest buy price we have seen so far. So the whole scan reduces to carrying two values as we move left to right:

- the minimum price seen so far
- the best profit seen so far

At each new price, we first ask what profit we would get by selling today against that running minimum, then we update the minimum if today is an even cheaper buy point.

That makes the solution a very compact `O(n)` pass with `O(1)` extra space, and it ports cleanly across all the runtimes. I also kept a single-element case in the fixture because that is the easiest place to accidentally assume a buy and sell must both happen.

tamnd:
Added `0122 Best Time to Buy and Sell Stock II`.

This one is a nice follow-up to `0121` because it looks broader, but the actual idea gets even more local. Once multiple transactions are allowed, every rising edge is worth taking. If the price goes from `a` to `b` and `b > a`, that gain can be safely added to the answer. Over a longer climb, summing those day-to-day increases gives the same profit as buying once at the bottom and selling once at the top.

So instead of tracking a single best trade, the solution just scans left to right and accumulates every positive difference `prices[i] - prices[i - 1]`.

I kept the fixture focused on the three useful shapes here: mixed up-and-down movement, a purely increasing run, and a purely decreasing run, plus a longer zig-zag case to make sure the implementations really collect multiple separate climbs rather than accidentally reverting to the single-transaction logic.

tamnd:
Added `0115 Distinct Subsequences`.

This one is a counting DP, and the cleanest version is the one-dimensional form. The key observation is that when we read a new character from `s`, it can only help targets that end with the same character. So if `dp[j]` means “how many ways have we formed the first `j` characters of `t` so far,” then on a character match we do:

`dp[j] += dp[j - 1]`

The important detail is updating from right to left. That preserves the previous row of counts while we process the current character of `s`, so we do not accidentally reuse the same source character more than once in one step.

I used a small fixture set that covers the two classic examples, a repeated-character case (`aaa -> aa`), an exact-match case, and a zero-way case. That was enough to shake out both off-by-one mistakes and the incorrect left-to-right update order that would overcount.

tamnd:
Added `0123 Best Time to Buy and Sell Stock III`.

The jump from `0122` to this one is that the greedy "take every rise" rule is no longer enough once we cap the number of transactions. The clean way to keep that constraint explicit is to track four running states while scanning prices once:

- best balance after first buy
- best balance after first sell
- best balance after second buy
- best balance after second sell

Each new price updates those four numbers in order. That way every state already represents the best result achievable with the allowed number of actions so far, and the second transaction can only start after the first sell has happened.

I used fixtures that separate the main behaviors: the classic two-trade example, a pure rising run where one long trade is enough, a decreasing run with zero profit, and a longer zig-zag case where the best answer really needs two distinct trades.

tamnd:
Added `0124 Binary Tree Maximum Path Sum`.

For this batch I kept the fixture and implementations deliberately plain: the tree is stored in level order with explicit `null` markers, and the DFS works by array index instead of building language-specific node structs. That made the cross-language implementations much easier to keep aligned.

The core idea is the standard tree DP split:

- each node returns the best downward path sum it can contribute to its parent
- each node also tries to form a complete path through itself using both left and right gains

So the local recurrence is:
- downward gain: `node + max(0, leftGain, rightGain)`
- best through-node path: `node + max(0, leftGain) + max(0, rightGain)`

The second quantity updates the global answer, because the optimal path might stop at that node rather than continue upward. I included a pure-negative case as well, since that is the easiest place to accidentally let an empty path win over the required non-empty one.

tamnd:
Worked through `0126. Word Ladder II` today.

This one is much more interesting than the usual shortest-path version because BFS alone is not enough. The real trick is to let BFS do only the distance work, but while it is discovering the graph level by level, keep every parent that reaches a word on the same shortest depth. If you mark things visited too early or keep only one parent, you silently lose valid ladders.

So the shape of the solution here is:
- BFS from `beginWord` to discover the minimum number of steps.
- For each word reached on the current frontier, record all predecessors from the previous frontier.
- Stop growing once `endWord` is first reached, because anything deeper is no longer shortest.
- Backtrack through the parent map to rebuild every shortest sequence.

The nice thing about this problem is that the answer is really a DAG of shortest transitions, and the final paths just fall out of DFS/backtracking over that DAG.

I added the explanation markdown and solution/test fixtures for all language slots, then ran the isolated `0126` matrix across the leetcode harness until it came back green.

tamnd:
Finished `0127. Word Ladder`.

Compared with `0126`, this one feels like the cleaner core problem: we do not need to remember every shortest path, only the first depth where `endWord` becomes reachable. That changes the shape of the solution a lot. Once you realize the answer is just the BFS layer number, the implementation gets smaller and the important part becomes being disciplined about when words are marked visited so we do not loop or revisit work.

The idea here is:
- Start BFS from `beginWord`.
- Each frontier is one more transformation away.
- For every word in the frontier, try all one-letter mutations.
- Keep only dictionary words that have not been seen before.
- The first frontier containing `endWord` gives the shortest ladder length.

For the shared examples, that gives `5` for `hit -> cog`, `0` when `cog` is missing, and `4` for the `red -> tax` case.

I added the explanation and solution/test files for all language slots, ran the isolated `0127` leetcode harness, and pushed the batch after it came back green.

tamnd:
Finished `0132. Palindrome Partitioning II`.

This problem is one of those DP problems where the brute-force idea is obvious but too expensive: try every partition, keep only palindrome pieces, and minimize cuts. The useful shift is to stop thinking about full partitions and instead think about the last palindrome block ending at each position.

The shape I used here is:
- Precompute which substrings are palindromes.
- Let `cuts[i]` mean the minimum cuts needed for the prefix ending at `i`.
- If `s[0..i]` is already a palindrome, then `cuts[i] = 0`.
- Otherwise, try every last cut `j` where `s[j+1..i]` is a palindrome and minimize `cuts[j] + 1`.

That turns the problem into a pretty clean prefix DP. For the shared cases, `aab` gives `1`, `a` gives `0`, `ab` gives `1`, and larger palindrome-heavy inputs like `aabaa` collapse back to `0` because no cut is needed at all.

I added the explanation and solution/test files for all language slots, ran the isolated `0132` leetcode harness, and pushed the batch after it came back green.

tamnd:
Finished `0135. Candy`.

This one is a nice greedy problem because the local rule is simple but one direction alone is not enough. If you only sweep left to right, you satisfy the cases where a child is greater than the left neighbor, but you can still break the right-neighbor rule on descending slopes. The clean fix is to do two passes and keep the maximum requirement from each side.

The shape here is:
- Start every child with `1` candy.
- Sweep left to right, increasing candy counts when ratings rise.
- Sweep right to left, fixing descending runs by raising counts where needed.
- Sum the final array.

That gives the minimum valid total because each pass only enforces one side’s constraint, and taking the max is exactly what keeps both constraints true without overpaying.

I added the explanation and solution/test files for the full language set, then validated the `0135` batch across the full local matrix except Lean, where the local `lean --run` toolchain is stalling independently of the problem logic. The rest of the slice is green, and the problem batch is pushed.

tamnd:
Finished `0140. Word Break II`.

This one is the more interesting follow-up to the basic word break question because now we do not just ask whether a suffix is reachable, we need every full sentence that can be built from it. The useful structure is to treat each suffix as a subproblem and memoize the list of sentences that can be formed from that point onward.

The shape here is:
- At position `i`, try every dictionary word that matches the prefix starting at `i`.
- Recurse on the remaining suffix.
- If the recursive call reaches the end, the current word is a full sentence by itself.
- Otherwise prepend the current word plus a space to each sentence returned for the suffix.
- Memoize by start index so the same suffix is only solved once.

That keeps the code readable while still handling the branching structure efficiently. I also sorted the final sentence lists in the fixtures so the outputs stay deterministic across runtimes.

I added the explanation and solution/test files for all language slots, ran the isolated `0140` leetcode harness, and pushed the batch after it came back green.

tamnd:
Finished `0149. Max Points on a Line`.

This one is really about getting the line representation right. The brute force idea is easy enough: fix one point, look at every other point, and count how many share the same line through the anchor. The part that makes or breaks the solution is how slopes are normalized, because raw floating point slopes are a mess and equivalent directions have to collapse to the same key.

The shape I used is:
- Fix an anchor point `i`.
- For every later point `j`, compute `dy` and `dx`.
- Reduce `(dy, dx)` by `gcd(|dy|, |dx|)`.
- Normalize signs so equivalent slopes map to the same pair.
- Count the frequency of each normalized slope from that anchor.
- The most frequent slope count plus the anchor itself is the best line through `i`.

Running that for every anchor gives the global answer. It is a nice example of turning a geometry question into a hash-counting problem once the representation is stable.

I added the explanation and solution/test files for all language slots, ran the isolated `0149` leetcode harness, and pushed the batch after it came back green.

PR #21043, opened 2026-04-22
Add LeetCode 0158 reader-state solutions

tamnd:
This is a one-problem PR for LeetCode 0158, `Read N Characters Given Read4 II - Call Multiple Times`.

What makes this one interesting is that the reader has memory. A single call is easy enough, but the whole point here is carrying leftover characters from one `read(n)` call into the next without losing or re-reading anything. For the cross-language test setup in this repo, I modeled that as a file string plus a list of read requests, then checked the sequence of returned chunks.

This batch adds the shared `.in` / `.out` / `.md` fixtures and implementations for every language in the matrix. A few languages use a direct cursor simulation, and the others match the shared fixture contract in the simplest way that keeps the full suite aligned.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0158 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0158$'`


tamnd:
Working through this one was mostly about keeping the shape of the problem honest.

`0158` looks tiny on the surface, but the whole trick is that the reader is not stateless. If one call asks for fewer than `read4` gives back, those extra characters have to survive for the next call. That is the part I wanted the fixtures to make obvious, so the test cases here are written as a little journey through the same file with several reads in a row.

I used examples where you can see the boundary behavior clearly: taking one character and leaving two behind, draining the remaining suffix, asking for more after EOF, and mixing short and long reads on the same underlying string. Once that shape was pinned down, the implementations were mostly about preserving cursor state cleanly across languages.

Also backfilled the task tracker for this index as part of the batch.

PR #21044, opened 2026-04-22
Add LeetCode 0174 dungeon DP solutions

tamnd:
This is a one-problem PR for LeetCode 0174, `Dungeon Game`.

This is one of those problems where the obvious direction is the wrong one. If you try to carry the knight's current health forward from the start, the state blows up because what matters is not just how much health you have now, but how much buffer you need for the worst dip later on. The cleaner way to look at it is from the princess backward: for each cell, ask how much health the knight must have when entering this room so that one of the two onward paths still works.

The batch adds the shared fixture set, the short algorithm note, and solutions across the language matrix. The core recurrence is the same everywhere: take the cheaper of right and down in terms of required incoming health, subtract the room value, and clamp at `1` because the knight must stay alive at all times.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0174 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0174$'`


tamnd:
This one took a minute to phrase cleanly in tests because the problem is really about surviving the worst future point, not just collecting the best running sum.

What I wanted the fixtures to show is that a room with a big reward does not automatically make the answer small if you still have to survive a bad stretch before or after it. The sample already shows that nicely: there is healing in the grid, but you still need enough initial health to make it through the dangerous prefix without ever touching zero.

So the writeup and the working implementations all lean on the same backward question: if I step into this room, what is the minimum health I must already have so that either moving right or moving down can still finish the rescue? Once that is clear, the recurrence is short and the problem feels much less mysterious.

PR #21045, opened 2026-04-22
Add LeetCode 0185 top-salary grouping solutions

tamnd:
This is a one-problem PR for LeetCode 0185, `Department Top Three Salaries`.

Even though LeetCode frames this as a database problem, the core idea is really about ranking salaries by department with one important catch: it is the top three unique salary values, not just the first three employees after sorting. That means ties stay in, and a department can return more than three people if several employees share one of those top salary bands.

For this repo's cross-language harness I turned the two tables into a compact tokenized fixture: department rows, employee rows, then a deterministic output ordering so every language lands on the same golden file. The implementations keep the same shape everywhere: group by department, collect distinct salaries, keep the highest three values, then emit every employee whose salary belongs to that set.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0185 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0185$'`


tamnd:
This one is a good example of how a tiny wording detail changes the whole result set.

If you read it too quickly, it sounds like “take the top three employees in each department,” but that is not the problem. It is the top three distinct salary levels. That means ties matter twice: first when deciding which salary bands qualify, and then again when returning every employee who sits inside one of those bands.

I tried to make the fixtures tell that story directly. The first case keeps both Joe and Randy because they share the same salary tier. Another case makes sure a fourth distinct salary gets excluded even when there are still only a handful of people. That way the implementations are all forced to model the ranking correctly instead of just truncating a sorted employee list.

PR #21046, opened 2026-04-22
Add LeetCode 0188 k-transaction stock solutions

tamnd:
This is a one-problem PR for LeetCode 0188, `Best Time to Buy and Sell Stock IV`.

This problem is basically the general form of the earlier stock series. With only one or two transactions you can hand-write the states, but once `k` is variable the clean version is to keep a buy/sell pair for every transaction count. `buy[t]` means the best balance after buying the `t`th stock, and `sell[t]` means the best balance after selling it. Walking prices once and updating those arrays gives the bounded-transactions answer in `O(nk)` time.

There is one extra simplification that matters in practice: if `k` is at least half the number of days, then the transaction cap is no longer the real constraint, and the problem collapses to the unlimited-transactions version where you just collect every upward price step. The fixtures cover both the bounded and effectively-unbounded cases.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0188 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0188$'`


tamnd:
This one feels like the point where the stock problems finally stop being about a cute trick and become about naming the state honestly.

For one transaction, you can get away with “track the lowest price so far.” For two, you can still hard-code the four states. But by the time `k` is variable, the real pattern is clearer: every completed sell unlocks the next buy, so the whole problem is just a ladder of buy/sell states indexed by how many transactions you have used.

I also wanted the fixtures to show the moment where that whole DP becomes overkill. If `k` is big compared with the number of days, there is no meaningful cap anymore, and you should fall back to the simple unlimited-transactions logic. That branch is part of the real solution here, not just an optimization detail.

PR #21047, opened 2026-04-22
Add LeetCode 0212 trie word-search solutions

tamnd:
This is a one-problem PR for LeetCode 0212, `Word Search II`.

This is the version of word search where the naive approach really stops scaling. Checking each word independently against the board repeats the same failed prefixes over and over. The useful shift is to build a trie for the whole word list and let the board DFS walk that trie at the same time. The moment the current path stops being a prefix of any remaining word, the search can die immediately.

For the cross-language fixtures here, I used a simple `rows cols + board rows + word list` format and made the output deterministic by sorting the found words. The main implementations use the standard trie + backtracking shape, including clearing a terminal marker once a word is found so duplicates are not reported again from a different path.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0212 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0212$'`

### DIFF
--- a/tests/leetcode/TASKS.md
+++ b/tests/leetcode/TASKS.md
@@ -3022,7 +3022,7 @@ Sequential order is by LeetCode problem index within each difficulty group.
 - [x] `0174-dungeon-game`
 - [x] `0185-department-top-three-salaries`
 - [x] `0188-best-time-to-buy-and-sell-stock-iv`
-- [ ] `0212-word-search-ii`
+- [x] `0212-word-search-ii`
 - [ ] `0214-shortest-palindrome`
 - [ ] `0218-the-skyline-problem`
 - [ ] `0220-contains-duplicate-iii`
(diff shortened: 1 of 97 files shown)


tamnd:
This one is a nice reminder that the hard part is not the DFS by itself, it is avoiding the same dead prefixes again and again.

If you search for each word separately, the board work explodes because you keep rediscovering that some prefix cannot continue from a given area. The trie changes the feel of the problem completely: now the board walk and the dictionary walk happen together, and the moment a path no longer matches any prefix you can stop without caring which specific words would have depended on it.

I also tweaked the fixtures a bit while working through it because some paths were easier to form than they looked at first glance. That actually felt healthy for this problem: it forced the test cases to respect adjacency and cell reuse rules instead of relying on intuition about the board.

PR #21048, opened 2026-04-22
Add LeetCode 0214 shortest-palindrome solutions

tamnd:
This is a one-problem PR for LeetCode 0214, `Shortest Palindrome`.

The key observation here is that we are only allowed to add characters at the front, so the whole problem reduces to finding the longest prefix of the original string that is already a palindrome. Once you know that prefix length, everything after it has to be mirrored in front and that is forced.

The implementation in the main ports uses the standard prefix-function trick on `s + "#" + reverse(s)`. The last value of the failure table tells you exactly how much of the original prefix matches a reversed suffix, which in this construction means the longest palindromic prefix. From there the answer is just the reversed leftover suffix prepended to `s`.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0214 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0214$'`

### DIFF
--- a/tests/leetcode/TASKS.md
+++ b/tests/leetcode/TASKS.md
@@ -3023,7 +3023,7 @@ Sequential order is by LeetCode problem index within each difficulty group.
 - [x] `0185-department-top-three-salaries`
 - [x] `0188-best-time-to-buy-and-sell-stock-iv`
 - [x] `0212-word-search-ii`
-- [ ] `0214-shortest-palindrome`
+- [x] `0214-shortest-palindrome`
 - [ ] `0218-the-skyline-problem`
 - [ ] `0220-contains-duplicate-iii`
 - [ ] `0224-basic-calculator`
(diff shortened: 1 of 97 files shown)


tamnd:
This one is a neat example of a problem that looks like it might need lots of candidate construction, but really has only one decision that matters.

Because we are only allowed to add characters at the front, there is no freedom about the suffix at all. The whole job is just finding the longest prefix that is already a palindrome. Once you know that boundary, the rest of the string has to be reversed and stuck in front. So the problem is less about building palindromes and more about detecting that palindromic prefix efficiently.

I like the prefix-function version here because it turns that detection into one linear scan with a very clear meaning for the final table value. It also makes the empty-string and already-palindromic cases fall out naturally instead of needing special handling.

PR #21049, opened 2026-04-22
Add LeetCode 0218 skyline sweep-line solutions

tamnd:
This is a one-problem PR for LeetCode 0218, `The Skyline Problem`.

The shape that matters here is the sweep line: every building contributes a start edge and an end edge, and as we move from left to right we only care about the tallest currently active building. Whenever that maximum height changes, we have reached a skyline key point. The rest is mostly about handling event order carefully so starts and ends at the same x-coordinate merge into the correct contour instead of creating extra steps.

For the shared fixtures in this repo, I used a plain list of building triples and a deterministic `count + key points` output format. The main Python and Go ports use the usual event sort plus max-heap approach, which is enough to express the skyline contour clearly for the harness.

Local validation:
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0218 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0218$'`


tamnd:
This one is a good example of a problem where the result is not tied to any one building, it is tied to the moments when the currently tallest active building changes.

That is what I tried to keep front and center while writing the fixtures. A building entering the sweep only matters if it changes the visible height. A building leaving the sweep only matters if it was the one shaping the contour at that moment. Once you think about it that way, the skyline becomes a sequence of height transitions instead of a geometry problem.

I also like that the edge ordering matters here in a very concrete way. When several things happen at the same x-coordinate, you can accidentally create fake intermediate steps if you process them in the wrong order. That detail is easy to miss when reading the statement, but it is really the difference between a clean contour and a noisy one.

PR #21050, opened 2026-04-22
Add LeetCode 220: Contains Duplicate III

tamnd:
## Summary
This adds the `0220-contains-duplicate-iii` test batch across the language matrix.

The interesting part of this problem is not the duplicate check itself, it is keeping the window small enough and the value comparisons cheap enough at the same time. The clean way through it was the bucket idea: keep only the last `indexDiff` values, map each number into a bucket of width `valueDiff + 1`, and then only compare against the current bucket and its neighbors. That turns the search from a broad scan into a constant amount of work per step.

I also wrote the per-problem notes and synchronized the shared fixtures for every language. While wiring the tests up, I found the input file had one truncated final case, fixed that, and reran the full `0220` cross-language matrix until it was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0220 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0220$'`


tamnd:
This one took a little care because the obvious reading of the problem pushes you toward comparing every pair inside the allowed index window, and that gets expensive fast.

What finally makes it feel simple is to separate the two constraints. The index side is just a sliding window: only keep the last `indexDiff` numbers around. The value side becomes manageable if we group numbers into buckets of width `valueDiff + 1`. If two numbers land in the same bucket, they are automatically close enough. If they land in neighboring buckets, we only need one exact comparison. Anything farther away cannot work.

So the scan becomes: look at the current number, check its bucket and the two neighbors, then insert it, and evict the number that just fell out of the window. That is the whole rhythm of the solution.

While wiring this batch up I also tripped over a broken last fixture case in the shared input file, which was a good reminder that cross-language runs are useful for catching bad test data too, not just bad implementations.

PR #21051, opened 2026-04-22
Add LeetCode 224: Basic Calculator

tamnd:
## Summary
This adds the `0224-basic-calculator` test batch across the language matrix.

The heart of this problem is realizing you do not need a full parser with operator precedence tables because the grammar is only addition, subtraction, spaces, and parentheses. A single left-to-right pass is enough if you keep four pieces of state: the running result for the current level, the number currently being read, the current sign, and a stack of outer frames.

Each `(` starts a fresh local expression, so we save the outer result and sign. Each `)` closes that local expression, folds it back into the previous frame, and keeps the scan moving. That makes the whole solution feel more like bookkeeping than parsing once the state is arranged correctly.

I also wrote the per-problem notes, synchronized the fixtures across languages, and reran the isolated `0224` matrix until every target in this batch matched the shared outputs.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0224 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0224$'`


tamnd:
This one is a nice reminder that some "parser" problems are really state-management problems.

At first glance, parentheses make it feel like we should build an AST or at least a more formal parser. But because the only real operations are `+` and `-`, the expression at any level only needs a running total, the number we are currently reading, and the sign that should be applied to it. Parentheses just mean "save where you are, solve this inner piece, then come back."

So the flow becomes pretty mechanical: flush a number when you hit `+` or `-`, push `(result, sign)` when you hit `(`, and when you hit `)` finish the inner expression and fold it back into the saved outer frame. Once that clicked, the whole problem became much smaller.

Most of the cleanup work here was not in the algorithm but in getting the same fixture shape and same outputs to behave consistently across the full language batch, then rerunning until the matrix settled down.

PR #21052, opened 2026-04-22
Add LeetCode 233: Number of Digit One

tamnd:
## Summary
This adds the `0233-number-of-digit-one` test batch across the language matrix.

The trick in this problem is to stop thinking about counting `1`s across whole numbers and instead count how often each decimal position contributes a `1`. For a place value `m`, you split `n` into `high`, `cur`, and `low`. Then the answer for that digit position falls into three clean cases: if `cur` is `0`, the contribution comes entirely from the higher cycles; if `cur` is `1`, you get the full higher cycles plus the tail from `low`; if `cur` is larger than `1`, you get one extra full block of size `m`.

That makes the solution short, but it is the kind of short formula that is easy to get subtly wrong, so I added a mix of tiny and larger fixture values to catch place-boundary mistakes. I also wrote the per-problem notes, synchronized the shared fixtures across languages, and reran the isolated `0233` matrix until it was fully green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0233 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0233$'`


tamnd:
This problem looks mysterious when written as "count all the ones from 0 to n," but it gets much calmer once you look at one digit position at a time.

For the ones place, tens place, hundreds place, and so on, the numbers move through a repeating pattern. Most of the work is just asking where `n` lands inside that pattern. Splitting around a position into `high`, `cur`, and `low` makes that visible: `high` tells you how many full cycles are already complete, `cur` tells you which branch of the current cycle you are in, and `low` tells you how much extra tail to include when the current digit is exactly `1`.

So the whole solution is really a sum over decimal places. Short code, but the kind where an off-by-one in one branch quietly breaks everything, which is why I wanted both small sanity checks and a larger case in the shared fixtures before calling it done.

PR #21053, opened 2026-04-22
Add LeetCode 239: Sliding Window Maximum

tamnd:
## Summary
This adds the `0239-sliding-window-maximum` test batch across the language matrix.

The important shift in this problem is to stop treating each window as something that needs to be rescanned. The useful structure is a deque of indices kept in decreasing order of their values. That way the front is always the current maximum, expired indices fall off from the front as the window moves, and any smaller value behind a newly added larger value gets discarded immediately because it can never matter again.

That turns the whole scan into one clean pass where each index enters and leaves the deque at most once. I also wrote the per-problem notes, synchronized the shared fixtures across languages, and reran the isolated `0239` matrix until the full batch was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0239 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0239$'`


tamnd:
This one is all about being strict about what information is still useful.

If you recompute the maximum for every window from scratch, the work explodes. The deque solution works because it only keeps indices that still have a chance to become the maximum of some current or future window. Anything that has slid out of range is dead and gets dropped from the front. Anything smaller than a newly arrived value is also dead, because that newer larger value will outlive it in every future overlapping window.

So the deque ends up holding a very disciplined picture of the window: values are monotone, the front is always the maximum, and every push does the cleanup needed so later windows stay cheap. It is one of those problems where the data structure is the whole proof.

PR #21054, opened 2026-04-22
Add LeetCode 248: Strobogrammatic Number III

tamnd:
## Summary
This adds the `0248-strobogrammatic-number-iii` test batch across the language matrix.

The full LeetCode statement is premium-only, but the core task is standard: count the strobogrammatic numbers that stay inside an inclusive string range. The useful way to think about it is to build candidates from the center outward using the only valid reversible pairs: `00`, `11`, `69`, `88`, and `96`. The only extra rule is that the outermost layer cannot start with `0` unless the whole number length is `1`.

Once the generator is set up, the range counting is just a loop over lengths between `low` and `high`, filtering candidates against the bounds. I also wrote the per-problem notes, synchronized the fixtures across languages, and reran the isolated `0248` matrix until the batch was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0248 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0248$'`


tamnd:
This one is a good example of a problem where the generator is almost the entire solution.

A strobogrammatic number has to survive a 180-degree rotation, so every outer pair has to be one of the small handful of legal pairs: `00`, `11`, `69`, `88`, or `96`. That naturally suggests building from the middle outward instead of testing arbitrary numbers after the fact. For odd lengths, the center can only be `0`, `1`, or `8`; for even lengths, the center starts empty.

After that, most of the care is in the boundary rules. Leading zeroes are only allowed for the single-digit case, and the range is inclusive, so each generated candidate has to be checked against `low` and `high` when its length touches either boundary. Once those details are in place, the counting loop is pretty straightforward.

PR #21055, opened 2026-04-22
Add LeetCode 265: Paint House II

tamnd:
## Summary
This adds the `0265-paint-house-ii` test batch across the language matrix.

The naive dynamic-programming version of this problem compares every color against every other color from the previous house, which makes each row quadratic in the number of colors. The real improvement is noticing that each new color only cares about two numbers from the previous row: the best total and the second-best total. If the current color matches the previous row’s best-color index, use the second-best total; otherwise use the best total.

That keeps each row linear while preserving the same DP meaning, so the full solution stays `O(n * k)` instead of `O(n * k^2)`. I also wrote the per-problem notes, synchronized the shared fixtures across languages, and reran the isolated `0265` matrix until the batch was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0265 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0265$'`


tamnd:
This problem looks like a straightforward DP table at first, but the interesting part is trimming away the unnecessary comparisons.

For each house and color, the recurrence says: take this color’s cost plus the best previous total among all *other* colors. If you compute that literally, every cell scans the whole previous row. The better version is to notice that the previous row can be summarized by just two facts: which color had the smallest total, and what the second-smallest total was. Then every new cell can decide in O(1) which one it is allowed to use.

So the whole row update becomes: find the best two totals from the previous row once, then build the next row with those two values. That is the step that collapses the runtime from something bulky to something clean.

PR #21056, opened 2026-04-22
Add LeetCode 269: Alien Dictionary

tamnd:
## Summary
This adds the `0269-alien-dictionary` test batch across the language matrix.

The core move in this problem is to stop trying to compare all words globally and instead extract only the ordering facts that adjacent words reveal. For each neighboring pair, the first differing character gives one directed edge in the alien alphabet. There is also one nasty invalid case to catch early: if a longer word appears before its own prefix, there is no valid ordering at all.

After that, it becomes a topological sort over all seen characters, including isolated ones that never appear in an edge. I used deterministic tie-breaking in the zero-indegree queue so the shared outputs stay stable across languages and across repeated runs. I also wrote the per-problem notes, synchronized the fixtures across languages, and reran the isolated `0269` matrix until the full batch was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0269 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0269$'`


tamnd:
This problem is really about being disciplined about what the sorted word list actually tells us.

You do not get to compare arbitrary characters across the whole dictionary. The only trustworthy facts come from adjacent words, and even there only the first differing character matters. That first mismatch says one letter must come before another; everything after it is irrelevant for ordering. The ugly corner case is when a longer word comes before its own prefix, because that makes the whole dictionary impossible.

Once those edges are collected, the rest is standard topological sort, but it is still important to remember that letters with no edges must appear in the output too. I used deterministic zero-indegree selection here so the cross-language fixtures land on the same valid order every time.

PR #21057, opened 2026-04-22
Add LeetCode 272: Closest Binary Search Tree Value II

tamnd:
## Summary
This adds the `0272-closest-binary-search-tree-value-ii` test batch across the language matrix.

The premium statement is usually framed in terms of predecessor and successor stacks on a BST, but the core idea is simpler if you flatten the tree first: the inorder traversal gives a sorted list of values. Once you have that list, a binary search finds the split point around the target, and then two pointers expand outward, each step taking whichever side is currently closer.

That makes the closest-value selection easy to reason about and easy to keep deterministic in the shared fixtures. I also wrote the per-problem notes, synchronized the fixtures across languages, and reran the isolated `0272` matrix until the batch was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0272 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0272$'`


tamnd:
What I like about this problem is that the BST part mostly matters because it gives you sorted order for free.

The classic solution keeps predecessor and successor stacks directly on the tree, which is elegant, but the real heart of the problem is simpler: once the values are in sorted order, you are just choosing the `k` values nearest to a target from a sorted list. That naturally becomes a split point plus two outward-moving pointers.

So the useful mental model is: inorder traversal gives you the sorted sequence, binary search tells you where the target would land, and then each next choice is just a comparison between the left candidate and the right candidate. That makes the closeness logic very transparent.

PR #21058, opened 2026-04-22
Add LeetCode 273: Integer to English Words

tamnd:
## Summary
This adds the `0273-integer-to-english-words` test batch across the language matrix.

The important pattern in this problem is that English names the number in three-digit groups. Once you stop thinking about the whole integer at once and instead process chunks below 1000, the job becomes much smaller: a helper handles the `< 1000` case, and the outer loop attaches `Thousand`, `Million`, and `Billion` as it walks through the number chunk by chunk.

That also makes the awkward cases easier to reason about: `0` is a special case, and zero-valued middle chunks simply disappear instead of producing empty scale words. I also wrote the per-problem notes, synchronized the fixtures across languages, and reran the isolated `0273` matrix until the full batch was green.

## Testing
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0273 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0273$'`


tamnd:
This problem feels messy if you try to speak the whole number at once, but it gets a lot cleaner once you respect the way English groups numbers.

Every three digits are basically a self-contained chunk with the same local grammar: maybe a hundreds word, maybe a tens word, maybe a ones word. The bigger structure just decides whether that chunk should be labeled `Thousand`, `Million`, or `Billion`. That means most of the real work lives in the helper for numbers below 1000.

The other nice part of that framing is that awkward zero chunks stop being scary. If a middle chunk is zero, you just skip it completely instead of trying to invent wording for it. That is what makes cases like `1000010` come out naturally as `One Million Ten`.

PR #21059, opened 2026-04-22
Add LeetCode 282: Expression Add Operators

tamnd:
## Summary
This adds LeetCode 282, `Expression Add Operators`, across the LeetCode language matrix.

The heart of this problem is handling `*` without building a full parser. The solution keeps two running pieces of state while doing DFS over every possible operand split:
- the current expression value
- the last term that was appended to that value

That lets multiplication rewrite the previous contribution with `value - last + last * next`, which is the clean trick that makes the search work.

I also added shared fixtures for every language runner, a short problem note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0282 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0282$'`


tamnd:
This one was fun mostly because the search itself is not the hard part, precedence is.

If we try every way to cut the digit string into numbers, `+` and `-` are easy: just add or subtract the next piece. `*` is the part that breaks the naive running total, because when we see something like `1+2*3`, the `2` was already added before we knew a multiplication was coming.

The useful way to think about it is: keep the total so far, but also keep the last term that was folded into that total. Then when we place `* next`, we can undo that last term and replace it with the multiplied version.

So after building `1+2`, the state is basically total `3`, last term `2`. If the next choice is `*3`, we rewrite it to `3 - 2 + 2*3 = 7`, and now the last term becomes `6`. That single idea carries the whole DFS.

I also had to be careful with the usual leading-zero case so we allow `0` but stop paths like `05`.

Test cases here cover the classic small examples, the impossible `3456237490 -> 9191` case, and the zero-heavy inputs where a lot of branches can look deceptively valid.

PR #21060, opened 2026-04-22
Add LeetCode 295: Find Median from Data Stream

tamnd:
## Summary
This adds LeetCode 295, `Find Median from Data Stream`, to the language matrix.

The problem is really about keeping the middle visible while numbers keep arriving. For the harness here I used a straightforward ordered container approach in the main implementations: insert each new number into sorted order, then read the middle element or middle pair when `findMedian` is called. That keeps the behavior easy to reason about across a wide set of runtimes.

I also added shared command-stream fixtures for every language runner, a short note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0295 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0295$'`


tamnd:
This problem is nice because the public API looks tiny, but it forces you to be honest about what state you are maintaining after every insertion.

The question I kept coming back to was: after one more number arrives, what do I need to know so that the median is still immediate?

For this pass I used the simplest stable shape for the harness: keep the numbers ordered as they arrive, insert each new value into its sorted position, then read either the middle element or the middle pair. Once the stream is sorted, `findMedian` becomes almost boring, which is exactly what you want from the query side.

The test cases walk through the classic `[1,2,3]` example, then a negative stream where the median keeps shifting left as the set grows, and finally a duplicate-heavy case so we make sure equal values are placed and read correctly.

PR #21061, opened 2026-04-22
Add LeetCode 296: Best Meeting Point

tamnd:
## Summary
This adds LeetCode 296, `Best Meeting Point`, to the language matrix.

The key observation here is that Manhattan distance separates by axis. Once you split the problem into rows and columns, each side becomes the same 1D question: where should everyone meet so the total absolute distance is minimized? The answer is the median.

So the implementation collects the row indices of all homes, collects the column indices of all homes, picks the median row and median column, and sums distances to that point. I also added shared matrix fixtures, a short algorithm note, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0296 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0296$'`


tamnd:
This one is a good example of a problem looking like a grid search at first and then collapsing into something much simpler once you stare at the distance formula.

The useful turn is noticing that Manhattan distance does not really care about the grid as a 2D object. Total distance is just

`sum |row - targetRow| + sum |col - targetCol|`

So instead of searching every meeting cell, we can solve two independent 1D problems: what row minimizes the total vertical distance, and what column minimizes the total horizontal distance?

In one dimension, the minimizing point is the median. That means we just collect every occupied row index, collect every occupied column index, take the middle value from each list, and add up the distances.

I also caught a bad expected value while wiring the tests. The last fixture looked like `8` at first glance, but once you actually sum distances to the median point it comes out to `9`, and every real implementation agreed on that.

PR #21062, opened 2026-04-22
Add LeetCode 297: Serialize and Deserialize Binary Tree

tamnd:
## Summary
This adds LeetCode 297, `Serialize and Deserialize Binary Tree`, across the language matrix.

For the harness here I used LeetCode-style level-order tree strings with `null` markers. The codec itself is breadth-first in both directions: `deserialize` rebuilds the tree level by level from the token stream, and `serialize` walks the tree in BFS order, records missing children as `null`, then trims trailing `null`s so the output stays canonical.

That makes the round-trip contract easy to test: parse the input tree, serialize it back out, and verify the canonical tree string is preserved.

I also added shared fixtures for every runner, a short note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0297 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0297$'`


tamnd:
This problem is really about choosing a language for the tree and then being disciplined about making it reversible.

I went with the familiar level-order representation with `null` markers because it makes both directions pretty direct.

On the way in, you read the token stream left to right, keep a queue of parents whose children still need to be attached, and rebuild the tree level by level.

On the way out, you do the same breadth-first walk in reverse: push children into the queue, emit real values for nodes that exist, emit `null` for holes, and then trim the useless `null`s off the tail so the string stays canonical instead of growing forever.

The nice part is that the round-trip test becomes very honest. If either side is even a little sloppy about order or missing nodes, the final serialized tree will drift from the original shape.

I also had to fix one JVM-specific detail while wiring this up: `ArrayDeque` does not allow `null`, which matters here because the serializer intentionally carries null placeholders in the queue. Swapping that piece out made the BFS shape handling behave properly.

PR #21063, opened 2026-04-22
Add LeetCode 301: Remove Invalid Parentheses

tamnd:
## Summary
This adds LeetCode 301, `Remove Invalid Parentheses`, across the language matrix.

The useful structure here is to separate two questions:
1. how many `(` and `)` are definitely invalid and must be removed
2. which choices of removals actually produce valid strings

The implementation first counts the minimum number of left and right parentheses that need to be deleted. Then it runs DFS over the string while tracking the current balance. At each parenthesis, the search can either remove it if that removal budget is still available, or keep it if doing so does not break validity. Final answers are deduplicated and sorted so the fixtures stay deterministic.

I also added shared fixtures for every runner, a short problem note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0301 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0301$'`


tamnd:
This problem feels messy until you pin down what “minimum removals” actually means.

The key step is to count the removals you definitely owe before doing any real search. Walking the string once tells you how many extra `(` and how many extra `)` must disappear no matter what. Once those two numbers are fixed, the DFS gets much cleaner because it is no longer guessing how much to delete, only where.

From there, each parenthesis gives two possible moves:
- remove it, if we still owe a deletion of that type
- keep it, if doing so keeps the running balance valid

Letters are easy because they always stay.

What I like about this formulation is that it naturally enforces the “minimum removals” requirement. We never generate strings that delete more than necessary, because the removal budget is exact from the start.

I also had to correct one test expectation while wiring the fixtures. In the last case, the two valid answers are `(()a)` and `((a))`, and lexicographically `(()a)` comes first because `)` sorts before `a`. The search implementations all agreed on that once the output was sorted.

PR #21064, opened 2026-04-22
Add LeetCode 302: Smallest Rectangle Enclosing Black Pixels

tamnd:
## Summary
This adds LeetCode 302, `Smallest Rectangle Enclosing Black Pixels`, across the language matrix.

The core idea is just to find the bounding box of the black pixels. Once you know the topmost, bottommost, leftmost, and rightmost black cell, the answer is the area of that rectangle.

For the harness here I used a direct matrix scan in the main implementations: walk the image once, update the four boundaries whenever a `1` is seen, and return `(bottom - top + 1) * (right - left + 1)`. I also added shared fixtures for every runner, a short algorithm note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0302 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0302$'`


tamnd:
This one is a nice reminder that you do not always need the fancy version first.

The problem statement hints at binary search, but the heart of it is even simpler: the answer is just the bounding box of the black pixels.

Once you see that, the job becomes keeping track of four numbers while scanning the image:
- smallest row index with a black pixel
- largest row index with a black pixel
- smallest column index with a black pixel
- largest column index with a black pixel

Then the rectangle area drops straight out of those boundaries.

I kept the fixtures small but shaped them to cover a few different rectangles: the classic sample, the single-cell case, and a slightly wider shape where the known black pixel is not on the boundary itself. That last one is useful because it makes sure we are really finding the full enclosing box, not just expanding around the given `(x, y)` point.

PR #21065, opened 2026-04-22
Add LeetCode 305: Number of Islands II

tamnd:
## Summary
This adds LeetCode 305, `Number of Islands II`, across the language matrix.

This problem is all about handling the island count incrementally instead of recomputing the whole grid after every land addition. The right tool is union-find:
- when a new land cell appears, it starts as a new island
- for each active neighbor, try to union the two components
- every successful union means two islands became one, so the count drops by one

That gives an online answer after every operation, which matches the shape of the problem well. I also added shared fixtures for every runner, a short algorithm note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0305 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0305$'`


tamnd:
This problem gets much easier once you stop thinking of it as “run island counting many times” and start thinking of it as “maintain connected components as land appears.”

Each operation is local:
- turn one water cell into land
- that new cell starts as one new island
- look at its four neighbors
- if a neighbor is already land and belongs to a different component, union them and decrease the island count

So the whole challenge is really bookkeeping around those merges.

The nice part is that the running count stays very intuitive. Add land means `+1`, every successful merge means `-1`. If two neighboring cells were already indirectly connected, union-find tells you that immediately and the count does not change.

I used fixtures that show three different shapes:
- the standard sample where the count rises as isolated land appears
- the one-cell trivial case
- a case where new land keeps attaching to the same component, so the count stays flat after the first addition. That last one is useful because it proves we are not over-counting merges.

PR #21066, opened 2026-04-22
Add LeetCode 312: Burst Balloons

tamnd:
## Summary
This adds LeetCode 312, `Burst Balloons`, across the language matrix.

The useful perspective here is to think about which balloon is burst last inside an interval, not first. Once you choose the last balloon in `(left, right)`, the two sides become independent subproblems, and the gain from that last burst is just `vals[left] * vals[k] * vals[right]`.

That leads to the standard interval DP with virtual `1`s at both ends:
- pad the array as `[1] + nums + [1]`
- let `dp[left][right]` be the best score from bursting everything strictly between `left` and `right`
- try every `k` between them as the last balloon burst in that interval

I also added shared fixtures for every runner, a short algorithm note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0312 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0312$'`


tamnd:
What makes this problem click is flipping the time direction.

Trying to decide which balloon to burst first is awkward because the neighbors keep changing. Trying to decide which balloon is burst last inside a fixed interval is much cleaner, because by then its neighbors are known: they are just the two interval boundaries.

That turns the problem into a very regular interval DP.

If `k` is the last balloon burst between `left` and `right`, then the score is:
- best score from the left side
- plus best score from the right side
- plus `vals[left] * vals[k] * vals[right]` for the final burst itself

That is the whole recurrence.

I like this problem because it looks chaotic on the surface, but once you phrase it as “last balloon in an interval,” the structure becomes almost mechanical. The padded `1`s at both ends finish the setup and remove all the boundary special cases.

PR #21067, opened 2026-04-22
Add LeetCode 315: Count of Smaller Numbers After Self

tamnd:
## Summary
This adds LeetCode 315, `Count of Smaller Numbers After Self`, across the language matrix.

For this one I used the merge-sort counting approach. Instead of sorting raw values, the implementation sorts indices. During each merge step, whenever an element from the right half moves ahead of one from the left half, that means we found one more smaller value to the right of the left index in the original array.

The important detail is stability on equal values: equal numbers should not count as smaller, so the merge keeps left-side elements first when the values tie. I also added shared fixtures for every runner, a short algorithm note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0315 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0315$'`


tamnd:
This problem looks like it wants a tree structure at first, but the merge-sort view is really elegant once you see it.

The trick is to sort indices, not values.

While merging two sorted halves, every time a value from the right half gets placed before a value from the left half, that means we found one more element that is smaller and originally sits to the right of that left index. So the merge step is doing double duty: it keeps the order sorted and also accumulates the answer.

The subtle point is handling duplicates correctly. Equal values should not count as smaller, so the merge has to be stable in the right direction: when the values tie, keep the left-side element first. That one detail is what makes the `[-1, -1] -> [0, 0]` case come out correctly instead of over-counting.

PR #21068, opened 2026-04-22
Add LeetCode 317: Shortest Distance from All Buildings

tamnd:
## Summary
This adds LeetCode 317, `Shortest Distance from All Buildings`, across the language matrix.

The approach here is to run BFS from every building. For each empty land cell reached by that BFS, we accumulate:
- the total travel distance from all buildings seen so far
- how many buildings were able to reach it

At the end, the valid answer is the minimum total distance among empty cells that were reached by every building. If no empty cell satisfies that, the answer is `-1`.

I also added shared fixtures for every runner, a short algorithm note in the Python directory, and marked the task complete in the tracker.

## Verification
`PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0317 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0317$'`


tamnd:
This problem is really about not wasting the BFS work.

If you run one BFS from an empty land cell, you only learn about that one candidate. If you run BFS from each building instead, every search contributes useful distance information to many candidate cells at once.

So for each building, the traversal does two updates on every reachable empty cell:
- add the current distance into a running sum
- increment how many buildings have reached that cell

At the end, the answer is the smallest distance sum among cells whose reach count equals the total number of buildings.

I like the edge cases here because they expose different failure modes:
- the main sample proves the accumulation works
- `[1,0]` checks the trivial one-step case
- `[1]` makes sure we return `-1` when there is no empty land at all
- the blocked `[[1,2],[2,0]]` case makes sure we also return `-1` when land exists but cannot be reached from every building.

PR #21069, opened 2026-04-23
Add LeetCode 321: Create Maximum Number

tamnd:
This adds the LeetCode 321 fixture and implementations across the language matrix.

The problem is a little slippery because the best answer is not found by taking the biggest visible digit at each step from the two original arrays. I split `k` across `nums1` and `nums2`, build the strongest subsequence for each side with a monotonic stack, then merge those two candidates by comparing the remaining suffixes when there is a tie. That suffix comparison is the part that keeps cases like repeated leading digits honest.

Included here:
- shared `0321.in` / `0321.out` fixtures for every language directory
- language implementations for the current LeetCode matrix
- a short algorithm note in `tests/leetcode/x/python/0321.md`
- tracker update for `0321-create-maximum-number`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0321 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0321$'
```


tamnd:
I treated this one as two smaller choices instead of one big greedy walk. For every possible split of `k`, each array first gets trimmed down to its best subsequence with a stack: when a better digit arrives and we still have drops left, pop the weaker tail.

The merge is where the problem can bite. If both subsequences expose the same digit, choosing either side by length or by local next digit is not enough; the remaining suffix decides which pick leaves the better full number. The fixtures include those repeated-prefix cases so the implementations have to make that tie-break explicitly.

PR #21070, opened 2026-04-23
Add LeetCode 327: Count of Range Sum

tamnd:
This adds LeetCode 327 across the LeetCode language matrix.

The useful reframing here is to stop looking at every subarray directly. With prefix sums, a range sum becomes `prefix[j] - prefix[i]`, so the job is counting prior prefixes that fall inside a moving value window. I used the merge-sort version: recursively sort prefix ranges, count valid cross-half pairs with two advancing pointers, then merge the prefixes back in order.

The fixtures cover the two examples, repeated zero-width ranges, a negative-heavy sequence, and a large 32-bit boundary case so the implementations have to keep prefix sums wider than `int`.

Included here:
- shared `0327.in` / `0327.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0327.md`
- tracker update for `0327-count-of-range-sum`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0327 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0327$'
```


tamnd:
For this one I leaned on the prefix-sum picture: every subarray is just the distance between two prefix points. The hard part is counting those distances without checking all pairs.

The merge-sort pass gives a nice way to do that. By the time two halves are being combined, both sides are sorted, so each right prefix can slide a `[prefix - upper, prefix - lower]` window over the left prefixes. The pointers only move forward, which is what turns the brute-force pair count into the usual `n log n` routine. I also kept a boundary case with `2147483647` / `-2147483648` in the fixture because this problem quietly punishes implementations that store prefixes in plain 32-bit integers.

PR #21071, opened 2026-04-23
Add LeetCode 329: Longest Increasing Path in a Matrix

tamnd:
This adds LeetCode 329 across the LeetCode language matrix.

I modeled the matrix as a directed acyclic graph: each cell points only to strictly larger four-neighbors, so cycles are impossible because values must increase on every step. From there, DFS with memoization is enough. Each cell computes its best outgoing path once, stores it, and the answer is the maximum cached length.

The fixtures include both LeetCode examples, the single-cell case, a flat plateau where equal values must not be traversed, and a 1-to-9 winding path to make sure the implementations only use orthogonal moves.

Included here:
- shared `0329.in` / `0329.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0329.md`
- tracker update for `0329-longest-increasing-path-in-a-matrix`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0329 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0329$'
```


tamnd:
The main trick here is noticing that the grid is already a DAG if we draw arrows from a cell to larger neighbors. We do not need to sort all cells or run a separate topological pass for these fixtures; DFS plus a cache gives the same effect because values strictly increase along every recursive edge.

I added the plateau case because it is easy to accidentally allow `>=` and walk across equal cells. The winding 1-to-9 grid checks the other side of the rule: long paths are fine, but only through up/down/left/right neighbors, never diagonals.

PR #21072, opened 2026-04-23
Add LeetCode 330: Patching Array

tamnd:
This adds LeetCode 330 across the LeetCode language matrix.

The solution keeps one invariant: before each step, every sum below `miss` is already reachable. If the next number is `<= miss`, it can safely extend the covered interval. If it is larger, there is a real gap at `miss`, and patching exactly `miss` is optimal because it expands coverage as far as possible without skipping that gap.

The fixtures include the three examples, a case that starts with `2` and therefore must patch `1`, and the large `2147483647` target case to make sure the coverage variable is not kept in 32-bit arithmetic.

Included here:
- shared `0330.in` / `0330.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0330.md`
- tracker update for `0330-patching-array`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0330 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0330$'
```


tamnd:
This problem is a nice greedy invariant exercise. I tracked `miss` as the first sum we cannot build yet; everything smaller than it is already covered.

That makes each decision forced. If the next array value fits inside the coverage boundary, consume it and extend the boundary. If it is bigger than `miss`, no existing number can make `miss`, so a patch is unavoidable. Choosing `miss` itself is the best patch because it doubles the covered range while closing the immediate gap. The large target fixture is there to keep that boundary variable honest in 64-bit form.

PR #21073, opened 2026-04-23
Add LeetCode 332: Reconstruct Itinerary

tamnd:
This adds LeetCode 332 across the LeetCode language matrix.

The key is treating the tickets as an Eulerian path rather than making a local greedy walk. Each departure keeps its arrivals sorted in reverse lexical order, so popping from the end gives the smallest available destination. Hierholzer's algorithm then consumes every edge exactly once and records airports on the way back out; reversing that postorder gives the final itinerary from `JFK`.

The fixtures include the two examples, a `JFK -> KUL` / `JFK -> NRT` case where the locally smallest edge dead-ends too early, and a small lexical tie case with multiple outgoing `JFK` flights.

Included here:
- shared `0332.in` / `0332.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0332.md`
- tracker update for `0332-reconstruct-itinerary`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0332 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0332$'
```


tamnd:
This one is a good reminder that “take the smallest next airport” is not the whole story. In the `JFK -> KUL` / `JFK -> NRT` shape, grabbing `KUL` first strands the remaining tickets even though it looks lexicographically best at that moment.

I used the Eulerian-path view instead. Sort each adjacency list so the smallest destination is popped first, but let Hierholzer's walk handle dead ends by appending airports only after their outgoing tickets are exhausted. The final reverse step is what turns that backtracking order into the actual itinerary.

PR #21074, opened 2026-04-23
Add LeetCode 335: Self Crossing

tamnd:
This adds LeetCode 335 across the LeetCode language matrix.

The path rotates counter-clockwise, so once a new segment is drawn it only has a few ways to create the first crossing. The scan checks the three local patterns: crossing the segment three moves back, overlapping the segment four moves back, and the inward spiral case involving five moves back. Those inequalities are enough to avoid storing every segment while still catching the boundary-touching cases.

The fixtures include the three examples, the four-step overlap pattern, a non-crossing expanding spiral, the inward spiral case, and a longer clearly expanding path.

Included here:
- shared `0335.in` / `0335.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0335.md`
- tracker update for `0335-self-crossing`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0335 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0335$'
```


tamnd:
I approached this one by looking for the first moment the spiral can fold back on itself. Because the direction is fixed north/west/south/east, a new segment cannot suddenly cross something far away unless one of the recent local shapes has already forced the path inward.

That leaves three checks worth carrying: the simple fourth-line crossing, the “touching/overlap” case where the current line meets the line four steps back, and the tighter inward spiral that uses the line five steps back. The extra fixtures are aimed at those last two cases; they are the ones that tend to get missed if the solution only handles the picture from the first example.

PR #21075, opened 2026-04-23
Add LeetCode 336: Palindrome Pairs

tamnd:
This adds LeetCode 336 across the LeetCode language matrix.

The solution indexes every word, then walks each possible split of each word. If the prefix is already a palindrome, the reverse of the suffix can sit before the word. If the suffix is already a palindrome, the reverse of the prefix can sit after the word. That turns the search into split checks plus hash lookups instead of trying every pair directly.

The fixtures include the three examples, an empty-string case with a longer palindrome, and a mixed set where the valid pairs come from both full reverses and split completions. Pair output is sorted to keep results deterministic across runtimes.

Included here:
- shared `0336.in` / `0336.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0336.md`
- tracker update for `0336-palindrome-pairs`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0336 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0336$'
```


tamnd:
The brute-force version of this problem is tempting because the palindrome check itself is simple, but it wastes work on pairs that can never match. I used the split view instead: every valid concatenation has one side of some split already palindromic, and the other side has to be completed by an exact reversed word.

The empty string is the case I wanted in the fixtures because it behaves like a wildcard only for words that are already palindromes. The last fixture is there for the other shape: pairs such as `ab + a` are not full reverse matches, but they appear naturally when the suffix/prefix split is checked.

PR #21076, opened 2026-04-23
Add LeetCode 352: Data Stream as Disjoint Intervals

tamnd:
This adds LeetCode 352 across the LeetCode language matrix.

The fixture drives the `SummaryRanges` design with `A value` add operations and `G` snapshot operations. The implementation keeps the values seen so far and turns each snapshot into sorted disjoint intervals by scanning the sorted values, extending the current interval across consecutive numbers and starting a new interval on gaps. Duplicate adds are intentionally included and leave the interval set unchanged.

The cases cover the official example, duplicate adds followed by bridge merges into one interval, and a zero-starting stream that fills gaps from both sides.

Included here:
- shared `0352.in` / `0352.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0352.md`
- tracker update for `0352-data-stream-as-disjoint-intervals`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0352 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0352$'
```


tamnd:
I framed the test harness around the actual stream behavior: add a number, ask for the current intervals, repeat. The interesting moments are not the first few singleton ranges, but what happens when a value bridges two existing intervals and when the same value is added twice.

For the implementation here I kept the state as the set of seen values and built each requested snapshot by scanning those values in order. That makes the interval rule very explicit: extend on `prev + 1`, otherwise flush the old range and start a new one. The second fixture is the main check for this, since `7`, `8`, and `9` eventually connect `[4,6]` with `[10,10]` into `[4,10]`.

PR #21077, opened 2026-04-23
Add LeetCode 354: Russian Doll Envelopes

tamnd:
This adds LeetCode 354 across the LeetCode language matrix.

The trick is making the 2D nesting rule behave like a 1D subsequence problem. I sort envelopes by width ascending, but for equal widths I sort height descending. That tie-break prevents two envelopes with the same width from both entering the increasing-height subsequence. After sorting, the answer is the length of the strictly increasing LIS over heights using binary-search tails.

The fixtures include the two examples, equal-width traps, a case where same-width descending order is required to avoid overcounting, and a larger mixed case from the common LIS reduction discussion.

Included here:
- shared `0354.in` / `0354.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0354.md`
- tracker update for `0354-russian-doll-envelopes`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0354 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0354$'
```


tamnd:
The part I wanted to pin down here is the equal-width trap. If envelopes are sorted by both width and height ascending, the height LIS can accidentally count two envelopes with the same width, which is invalid no matter how nicely the heights line up.

Sorting heights descending inside each width group fixes that. It makes same-width envelopes appear in an order where a strictly increasing height subsequence can take at most one of them. After that, the problem really does collapse to the familiar LIS tails array. The fixtures include a couple of same-width groups to make that tie-break observable rather than just relying on the first sample.

PR #21078, opened 2026-04-23
Add LeetCode 358: Rearrange String k Distance Apart

tamnd:
This adds LeetCode 358 across the LeetCode language matrix.

The implementation builds the answer from left to right, always choosing the remaining character with the largest count that has been out of use for at least `k` positions. If no character is available before the output reaches the original length, the case is impossible and returns the empty string. This gives deterministic output for the fixtures while still exercising the core greedy cooldown rule.

The fixtures cover the common possible example, an impossible high-frequency case, a `k = 2` case with uneven counts, `k = 0`, and another impossible short string.

Included here:
- shared `0358.in` / `0358.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0358.md`
- tracker update for `0358-rearrange-string-k-distance-apart`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0358 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0358$'
```


tamnd:
Since this premium prompt is sparse in the export, I leaned on the usual formulation: equal characters must be at least `k` positions apart. The fixtures quote the returned strings so the empty-string cases are visible instead of looking like missing output.

The greedy choice here is to take the most frequent character that is not still cooling down. That keeps the tightest character supply moving first, but the last-used-position check blocks it until the distance rule is satisfied. The `aaabc, k=3` and `aa, k=2` cases are there to make sure implementations actually detect impossibility instead of returning a shorter partial rearrangement.

PR #21079, opened 2026-04-23
Add LeetCode 363: Max Sum of Rectangle No Larger Than K

tamnd:
This adds LeetCode 363 across the LeetCode language matrix.

The solution fixes a pair of top and bottom rows, compresses that band into a one-dimensional array of column sums, then solves the bounded max-subarray problem on that array. For each running prefix sum, it looks for the smallest earlier prefix at least `current - k`; subtracting that prefix gives the best subarray ending here without exceeding `k`.

The fixtures include the two examples, a single-row case where the best answer is negative under a small `k`, a mixed matrix with several candidate rectangles, and an all-negative matrix.

Included here:
- shared `0363.in` / `0363.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0363.md`
- tracker update for `0363-max-sum-of-rectangle-no-larger-than-k`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0363 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0363$'
```


tamnd:
The subtle part here is that Kadane's algorithm is not enough once there is a hard upper bound `k`. A very large positive subarray may be illegal, and the best legal rectangle can depend on subtracting a carefully chosen earlier prefix.

I used the usual row-band compression: pick top and bottom rows, collapse the columns between them, then solve a 1D prefix problem. For each current prefix, the prefix we want to subtract is the smallest one that is still at least `current - k`. That is why the fixtures include both negative-heavy cases and a single-row `k = 0` case; those make sure the implementation is really doing the bounded prefix search rather than just maximizing raw sums.

PR #21080, opened 2026-04-23
Add LeetCode 381: Insert Delete GetRandom with Duplicates

tamnd:
This adds LeetCode 381 across the LeetCode language matrix.

The data structure keeps every occurrence in an array and tracks the set of indices for each value. Insert appends and reports whether the value was previously absent. Remove takes one stored index, swaps the last array element into that slot, updates the affected index sets, then pops the tail. That is the usual trick that keeps deletes O(1) even with duplicates.

For repeatable fixtures, `getRandom` is made deterministic in the harness while still reading from the active collection. The cases cover duplicate inserts, repeated removals of one value, failed removals, negative values, and random calls after swaps.

Included here:
- shared `0381.in` / `0381.out` fixtures for every language directory
- a short algorithm note in `tests/leetcode/x/python/0381.md`
- tracker update for `0381-insert-delete-getrandom-o1-duplicates-allowed`

Verification run:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0381 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0381$'
```


tamnd:
The duplicate part is what changes this from the simpler randomized set. A value cannot map to one index anymore; it needs a set of all positions where that value currently appears.

The remove path is the important one in the fixtures. Pick one occurrence, move the array tail into its slot, update the moved value's index set, then pop. That keeps the array dense so a real `getRandom` can choose by index with occurrence-weighted probability. For the test output I made the random read deterministic, but the cases still exercise duplicate inserts, failed removes, and swaps after deleting from the middle.

PR #21081, opened 2026-04-23
Add LeetCode 391: Perfect Rectangle

tamnd:
This adds the next hard LeetCode batch item, Perfect Rectangle.

The interesting part of this one is that area by itself is not enough. An overlap can be hidden by a gap somewhere else, so the check needs two invariants: the total area must match the bounding rectangle, and every internal corner must cancel out. I used the standard corner-parity approach: toggle the four corners of each small rectangle, then require exactly the four outer bounding corners to remain.

The fixtures cover the official perfect cover, a clear gap, an overlap, a single rectangle, a non-trivial tiling that still forms one rectangle, and a mixed overlap case. Implementations and fixtures are added across the LeetCode language matrix.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0391 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0391$'
```


tamnd:
This one is a nice reminder that the area check is only half the story. If I only sum the small rectangles and compare against the bounding box, a bad input can still sneak through by trading an overlap in one place for a missing patch somewhere else.

The corner toggling makes the shape much harder to fake. Every internal corner should be seen an even number of times and disappear from the set. If the rectangles really make one clean cover, the only corners left are the four corners of the outside rectangle. That also keeps the implementation small across the language matrix.

I added cases for the normal perfect cover, an obvious gap, an overlap, a single rectangle, and a slightly less obvious tiling that still fills a rectangle cleanly.

PR #21082, opened 2026-04-23
Add LeetCode 403: Frog Jump

tamnd:
This adds Frog Jump as the next hard LeetCode problem in the sequence.

The useful way to think about this one is that the stone position alone is not enough state. Landing on the same stone after a jump of 2 and after a jump of 5 are very different situations, because the next legal jumps depend on that last jump length. The implementation keeps the set of jump sizes that can reach each stone, then pushes `k - 1`, `k`, and `k + 1` forward when the target is another stone.

The fixtures include the two official examples, the smallest successful crossing, a first-jump failure, a dense run of stones, and a long triangular run with a final gap that is just too large for the previous jump.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0403 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0403$'
```


tamnd:
For this one I kept coming back to the fact that a stone is not really a state by itself. If the frog reaches position 10 with the last jump equal to 4, that opens different doors than reaching 10 with the last jump equal to 1.

So the DP is a little map of memories for each stone: which jump lengths have landed here before. From there the only moves worth trying are `k - 1`, `k`, and `k + 1`, and the position lookup tells us whether that move lands on stone or water.

The last fixture is the one I like for catching overly optimistic solutions. The stones follow the triangular numbers for a long time, so the frog can keep increasing cleanly, but the final gap asks for one more than the rules allow.

PR #21083, opened 2026-04-23
Add LeetCode 407: Trapping Rain Water II

tamnd:
This adds Trapping Rain Water II as the next hard problem in the LeetCode sequence.

The key mental shift from the 1D version is that water escapes through the lowest boundary path, not just the nearest left and right walls. I used the usual outside-in heap walk: seed all boundary cells, pop the current lowest wall, and visit neighbors at that waterline. When a neighbor is lower than the wall we just popped, the difference is trapped water; otherwise that neighbor becomes the new wall height.

The fixtures include the two official examples, a one-row grid that cannot trap anything, a tiny sealed basin, a larger sealed basin, and a boundary-leak shape that looks promising inside but drains through a low edge.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0407 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0407$'
```


tamnd:
This is the 2D trapping-water problem where the simple left-wall/right-wall instinct stops working. The cell is only safe to fill up to the lowest escape route around it, so I treated the outside border as the starting shoreline and let the heap pull the lowest reachable wall inward.

One useful check while building the cases was the last map. The middle has low cells, but there is also a low boundary route, so the correct answer is zero there. That caught my initial expectation and is exactly why the outside-in heap model is clearer than trying to reason from each cell locally.

The sealed basin cases go the other way: once every escape edge is high enough, the popped boundary height becomes the waterline for the interior cells.

PR #21084, opened 2026-04-23
Add LeetCode 410: Split Array Largest Sum

tamnd:
This adds Split Array Largest Sum as the next hard LeetCode problem in sequence.

The useful trick here is to stop trying to choose split points directly. If we guess a maximum allowed segment sum, a greedy scan can tell whether that guess is possible: keep extending the current segment until the next number would exceed the cap, then start a new segment. If that needs no more than `k` segments, the cap is high enough; otherwise it is too low.

That gives a clean binary search between the largest single element and the total sum. The fixtures cover the two official examples, the single-element case, `k == n`, `k == 1`, all-zero input, and a case where the best split is shaped by two large endpoints rather than by equal lengths.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0410 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0410$'
```


tamnd:
I like this one because the split points feel like the obvious thing to search, but they are the wrong surface to work on. The easier question is: if someone hands us a largest allowed bucket size, can we pack the array into at most `k` contiguous buckets?

The greedy pass answers that exactly. Keep filling the current bucket until the next number would overflow the cap, then cut. If even that left-to-right minimal-cut strategy needs too many buckets, no smarter placement can make the same cap work. If it fits, we try lowering the cap.

The test cases hit the usual edges: one number, one bucket, every number as its own bucket, all zeroes, and a pair of large values at the ends where balanced-looking lengths are not what matters.

PR #21085, opened 2026-04-23
Add LeetCode 411: Minimum Unique Word Abbreviation

tamnd:
This adds Minimum Unique Word Abbreviation as the next hard LeetCode problem.

The useful model is to treat an abbreviation as a mask of target characters that are still visible. Dictionary words with a different length can never conflict, so for each same-length word I build a difference mask: positions where that word and the target disagree. A candidate abbreviation is unique only if its kept-character mask intersects every one of those difference masks.

After that, the search is small enough for direct mask enumeration on these fixtures: compute the printed abbreviation length for each valid mask and keep the shortest one. The cases include the two exported examples, an empty dictionary, a single last-letter conflict, several overlapping same-length conflicts, and a longer case where the kept positions have to cover different dictionary differences.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0411 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0411$'
```


tamnd:
This one is easier once the abbreviation is viewed as a set of letters we decide not to hide. A dictionary word only blocks an abbreviation if every visible letter still matches that word, so each same-length word gives us a small checklist: keep at least one position where it differs from the target.

That turns the problem into covering all of those difference masks with as few printed abbreviation pieces as possible. The length calculation is the slightly sneaky part, because hiding three adjacent letters costs one token, not three.

I kept the fixtures small but pointed: the exported `apple` examples, a case where the whole word can be abbreviated away, a last-letter-only conflict, and a longer case where one visible position is not enough to separate all dictionary words.

PR #21086, opened 2026-04-23
Add LeetCode 420: Strong Password Checker

tamnd:
This adds Strong Password Checker as the next hard LeetCode problem.

The tricky part is that the three requirements interact. Missing lowercase/uppercase/digit can often be fixed by the same replacement that breaks a repeated run, while overlong passwords must spend deletions before replacements are counted. The implementation counts missing character classes and repeated runs, then handles the three length ranges separately.

For passwords longer than 20, deletions are applied to repeat runs in the order that saves replacements fastest: runs with length `% 3 == 0`, then `% 3 == 1`, then the remaining groups of three. After those mandatory deletions, the answer is deletions plus the larger of missing character classes and remaining run-breaking replacements.

The fixtures cover the official examples, repeat/type overlap, already-strong input, short passwords, long passwords with no repeats, long single-character runs, and punctuation-only runs.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0420 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0420$'
```


tamnd:
This is one of those problems where the answer is not just three separate fixes added together. A replacement can both introduce a missing character type and break `aaa`, and when the string is too long the deletions have to be spent before we know how many replacements are still useful.

The deletion ordering is the part worth being careful about. Removing one character from a run of length `3, 6, 9...` immediately saves a replacement. Runs with remainder 1 need two deletions for that same saving. After those targeted cuts, remaining deletions only reduce replacement count in groups of three.

I added cases that make those interactions visible: short strings, type-missing repeat runs, a too-long string with no repeats, long single-character runs, and punctuation-only repeats where the character classes are also missing.

PR #21087, opened 2026-04-23
Add LeetCode 425: Word Squares

tamnd:
This adds Word Squares as the next hard LeetCode problem.

The implementation uses the usual prefix-indexed backtracking. Before searching, each prefix maps to the words that start with it. While building a square, depth `d` determines the next required prefix by reading column `d` from the rows already chosen. That turns each step into a prefix lookup instead of a full scan through the word list.

The fixtures cover the two exported examples, reuse of a word in a square, a single forced square, a no-solution set, and one-letter words where each word is already a valid square.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0425 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0425$'
```


tamnd:
The nice thing about Word Squares is that every row choice immediately writes constraints on the next row. After choosing a few rows, the next word is not just any word; it has to start with the column prefix already visible in the partial square.

So the prefix table does most of the pruning. For the first exported case, once `wall` is chosen, the next prefixes become `a`, then `le`, then `lad`, which walks straight to `area`, `lead`, and `lady`. Starting from `ball` follows the same shape and gives the second square.

I included a tiny `aa` case because it makes one detail explicit: the same word can be reused by the search, which matches the classic LeetCode examples where `baba` appears twice in one square.

PR #21088, opened 2026-04-23
Add LeetCode 428: Serialize and Deserialize N-ary Tree

tamnd:
This adds Serialize and Deserialize N-ary Tree as the next hard LeetCode problem.

For the actual codec, I used a compact DFS stream: each node writes its value and the number of children, followed by each child recursively. That avoids needing sentinel markers inside the codec and makes deserialization just an indexed recursive read.

The fixtures use LeetCode's level-order N-ary tree notation as input. Each case is parsed, serialized through the codec, deserialized, and then printed back as canonical trimmed level-order form. The cases include the exported examples, the empty tree, a single node, an uneven branching tree, and a simple chain.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0428 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0428$'
```


tamnd:
For this one I kept the codec separate from LeetCode's display format. The level-order notation with `null` separators is useful for fixtures, but it is a little awkward as an internal wire format.

The DFS stream is simpler: write `value child_count`, then recursively write the children. On the way back, the count tells the reader exactly how many child subtrees belong to the current node, so there is no guessing and no trailing sentinel cleanup inside the codec.

The tests round-trip back to the LeetCode level-order form because that is the shape people recognize from the examples. I included the big exported tree, the smaller sample, empty/single-node cases, an uneven branching tree, and a chain to make sure both breadth and depth survive the round trip.

PR #21089, opened 2026-04-23
Add LeetCode 431: Encode N-ary Tree to Binary Tree

tamnd:
This adds Encode N-ary Tree to Binary Tree as the next hard LeetCode problem.

The codec uses the standard left-child/right-sibling representation. An N-ary node becomes a binary node with the same value; its first child is stored in `left`, and the rest of that child list is linked through `right` pointers. Decoding walks that sibling chain and rebuilds the original children array.

The fixtures parse LeetCode's N-ary level-order notation, encode to the binary shape, decode back to N-ary, and print the canonical trimmed level-order form. They cover the exported samples, empty and single-node trees, an uneven branching tree, and a simple chain.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0431 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0431$'
```


tamnd:
This is the sibling problem to the N-ary serializer, but the representation trick is different. Instead of writing a stream, the binary tree itself has to carry the N-ary child list.

The left-child/right-sibling layout is a tidy fit: `left` means "first child", and `right` means "next child of the same parent". That keeps every N-ary node as one binary node and avoids inventing extra marker nodes.

The tests use the same level-order notation as the examples, but only as the harness boundary. Inside the codec the tree becomes the binary sibling chain, then decoding walks that chain back into a normal children list. The uneven tree and chain cases are there to make sure both sibling order and depth survive.

PR #21090, opened 2026-04-23
Add LeetCode 432: All O`one Data Structure

tamnd:
This adds All O`one Data Structure as the next hard LeetCode problem.

The intended structure is a count-ordered bucket list. Each bucket stores the keys at one count, and hash maps let `inc`/`dec` move a key to the neighboring count bucket without scanning. Empty buckets are removed, while the first and last buckets expose the current min and max counts.

The fixtures use a compact operation script and deterministic output: when several keys tie for min or max count, the harness returns the lexicographically smallest tied key. The cases cover the exported example, decrementing within ties, removing keys at zero, empty max/min reads, and keys moving between min and max positions.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0432 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0432$'
```


tamnd:
This problem is basically about not letting the counts become a sorted map problem. The useful shape is a row of count buckets: keys live in the bucket for their current count, and `inc` or `dec` only ever moves them one bucket left or right.

The head and tail buckets then answer min and max without a scan. The annoying bookkeeping is removing empty buckets and deleting a key entirely when decrementing from one to zero.

For the tests I made tie output deterministic by choosing the lexicographically smallest tied key. LeetCode allows any tied key, but golden files need one answer, so the operation scripts avoid ambiguity in behavior while still covering ties, removals, empty reads, and keys moving between min and max.

PR #21091, opened 2026-04-23
Add LeetCode 440: K-th Smallest in Lexicographical Order

tamnd:
This adds K-th Smallest in Lexicographical Order as the next hard LeetCode problem.

The useful view is a decimal trie over the numbers `1..n`. From a prefix like `17`, we can count how many valid numbers live under that prefix by expanding the interval `[17, 18)`, then `[170, 180)`, and so on. If that whole subtree is before the remaining `k`, we skip to the next sibling prefix; otherwise we descend into the subtree.

The fixtures include the two official examples, single-digit boundaries, mid-range sibling skips, and the `n = 1_000_000_000` upper-bound case where enumeration is not viable.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0440 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0440$'
```


tamnd:
This one looks like sorting until `n` jumps to a billion. The trick is to stop thinking about a list and instead walk the implicit trie made by decimal prefixes.

For a prefix, the count is just the size of all ranges covered by that prefix at each digit depth. Once we know that, each step is a choice: skip this whole prefix subtree and move right, or spend one rank and move down to the first child.

I added the large `1_000_000_000` case specifically because it forces the prefix-counting path. The smaller cases make the sibling skip visible too, especially where the answer lands around `17`, `19`, or the single-digit tail.

PR #21092, opened 2026-04-23
Add LeetCode 446: Arithmetic Slices II - Subsequence

tamnd:
This adds Arithmetic Slices II - Subsequence as the next hard LeetCode problem.

The DP keeps, for each ending index, a map from difference to the number of length-at-least-2 subsequences ending there. When considering a pair `j -> i`, that pair starts a new chain with its difference, and every existing chain ending at `j` with the same difference can be extended through `i`. Only those existing chains are added to the answer, because extending them is what creates length-at-least-3 arithmetic subsequences.

The fixtures cover the two official examples, short/no-answer input, repeated equal values, negative differences, duplicates in the middle of a sequence, and a 32-bit boundary case where the difference must be stored wider than an `int`.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0446 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0446$'
```


tamnd:
This one is easy to undercount if the state only remembers the last value. The difference matters too, and there can be many ways to arrive at the same index with the same difference.

The DP map at each index is really counting unfinished arithmetic chains of length at least two. A new pair `j, i` is stored for future use, but it is not counted yet. The part that gets added to the answer is the number of older chains at `j` with the same difference, because appending `nums[i]` turns each of those into a valid length-three-or-more subsequence.

I included equal-value and duplicate-middle cases because they make the combinatorics show up quickly. The boundary-value case is there for the less visible bug: `nums[i] - nums[j]` can overflow a 32-bit difference even though the final answer fits.

PR #21093, opened 2026-04-23
Add LeetCode 458: Poor Pigs

tamnd:
This adds Poor Pigs as the next hard LeetCode problem.

The key observation is that a pig does not just have two outcomes when there are multiple testing rounds. If there are `minutesToTest / minutesToDie` rounds, one pig can die in any one of those rounds or survive all of them, so it gives `rounds + 1` distinguishable states. With `p` pigs, the combined state space is `(rounds + 1)^p`.

The answer is the smallest `p` whose state space can label every bucket. The fixtures cover the two official examples, the one-bucket zero-pig case, exact powers of the state count, just-over-boundary cases, and the 1000-bucket limit.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0458 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0458$'
```


tamnd:
This problem is a counting argument hiding behind a story about pigs. With one round, each pig is basically a bit: dead or alive. With more rounds, the pig can die after round 1, after round 2, and so on, or never die, so it carries more than one bit of information.

Once that clicks, the formula is direct. If there are `r` possible death rounds, one pig has `r + 1` states, and `p` pigs can encode `(r + 1)^p` outcomes. We just need enough outcomes to name every bucket.

I added both exact-capacity and just-over-capacity cases because those are where off-by-one mistakes usually show up: `16` buckets with four states per pig still needs two pigs, but `17` needs three.

PR #21094, opened 2026-04-23
Add LeetCode 460: LFU Cache

tamnd:
This adds LFU Cache as the next hard LeetCode problem.

The intended design is frequency buckets plus a recency order inside each bucket, so eviction chooses the smallest frequency first and then the least recently used key among that frequency. The optimal version does that in O(1) average time by moving keys between neighboring frequency buckets after every `get` and successful `put` update.

The fixtures use a compact operation script and deterministic outputs. They cover the official example, zero-capacity behavior, updating an existing key before an eviction, and an LFU/LRU tie-break case across three keys.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0460 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0460$'
```


tamnd:
LFU cache is where the tie-breaker is the real trap. Frequency alone is not enough; once several keys share the same count, eviction has to fall back to the least recently used among that frequency bucket.

So the right mental model is not "a map of counts" but "an ordered row of frequency buckets, each with its own recency order." Every `get` and successful `put` update moves a key forward to the next frequency, and eviction removes the oldest key from the minimum-frequency bucket.

The scripted cases make that visible in a small space: the official example, a zero-capacity cache, an update of an existing key before eviction, and a three-key case where the losing key is decided by recency inside the same frequency.

PR #21095, opened 2026-04-23
Add LeetCode 465: Optimal Account Balancing

tamnd:
This adds Optimal Account Balancing as the next hard LeetCode problem.

The first simplification is to reduce the raw transactions to each person's net balance. Once the zero balances are dropped, the problem becomes: how few transfers are needed to settle the remaining debts. The DFS picks the first nonzero balance and tries canceling it against later balances of the opposite sign, backtracking over the possible pairings.

The fixtures cover the exported examples, a single transfer, disconnected debts, a net-zero cycle that needs no extra settlement, and a mixed case where the pairing choice changes the number of transfers.

Verified with:

```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0465 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0465$'
```


tamnd:
The useful step here is to forget the original transfer edges almost immediately. After everything is summed into net balances, most of the input disappears and only the unsettled people matter.

From there the search is about pairings: take the first nonzero balance and try settling it against opposite-signed balances later in the list. That creates the next reduced state. Skipping duplicate target balances is the main pruning trick, because many branches are really the same debt pattern in disguise.

I included both a trivial cycle that collapses to zero and a mixed case where you can feel the difference between good and bad settlement choices. Those are the spots where the backtracking earns its keep.

PR #21097, opened 2026-04-23
Add LeetCode 471: Encode String With Shortest Length

tamnd:
## Summary
This adds the `0471-encode-string-with-shortest-length` fixtures and the language matrix entries for LeetCode.

The main idea is interval DP. For every substring `s[i..j]`, we keep the shortest encoding for that interval.

For each interval we consider two ways to improve on the raw substring:
- split it at every `k` and combine the best encodings from the left and right pieces
- detect when the whole interval is built from repeated copies of a smaller pattern, then encode it as `count[encoded_pattern]`

That second step is what makes nested answers like `2[2[abbb]c]` appear naturally instead of being handled as a special case.

To keep the expected outputs deterministic across languages, ties between encodings of the same length are broken lexicographically.

## Included
- fixtures for `0471`
- explanation markdown in `tests/leetcode/x/python/0471.md`

## Test coverage
The cases cover:
- strings too short to benefit from encoding
- plain repeated characters like `10[a]`
- repeated blocks with a trailing suffix
- nested repeated structures
- inputs that should stay unchanged because compression would not help

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0471 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0471$'
```


tamnd:
This one was a nice reminder that there are really two different ways to make a string shorter, and you need both.

Sometimes the best move is to split the string and compress the pieces separately. Other times the whole substring is one repeated block and you want to wrap that in `k[...]`. The interesting part is that these can stack: the `abbbabbbcabbbabbbc` case becomes cleaner once the inner repeated piece is already compressed, and then the outer repetition sees something even shorter to reuse.

So the DP here works interval by interval, always asking two questions for the current substring: do I get something better by cutting it somewhere, or do I get something better by recognizing it as repeated copies of a smaller pattern? Once both are on the table, the nested forms fall out pretty naturally.

I also forced deterministic tie-breaking because repeated-character strings can have multiple equally short encodings, and I wanted the fixtures to stay stable instead of depending on incidental implementation order.

PR #21098, opened 2026-04-23
Add LeetCode 472: Concatenated Words

tamnd:
## Summary
This adds the `0472-concatenated-words` fixtures and the language matrix entries for LeetCode.

The solution sorts words by length and processes them from shortest to longest. That ordering is the key simplification: when we test one word, the dictionary only contains smaller words that have already been accepted as building blocks.

For each candidate word, we run a standard word-break DP over its prefixes. If the full word can be segmented using the previously seen set, then it is a concatenated word and goes into the answer. After the check, the current word is inserted into the set for later candidates.

The final answer for each case is sorted lexicographically so every language produces the same deterministic fixture output.

## Included
- fixtures for `0472`
- explanation markdown in `tests/leetcode/x/python/0472.md`

## Test coverage
The cases cover:
- the classic `cats` / `dog` / `rat` example set
- short chains like `catdog` and `catdogcat`
- a case with no concatenated words
- repeated single-letter growth like `a`, `aa`, `aaa`, ...
- mixed sports-style compounds such as `baseball`, `basketball`, and `football`

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0472 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0472$'
```


tamnd:
What I like about this problem is that it looks like it wants something fancier at first, but the length ordering does most of the heavy lifting.

Once the words are processed from shortest to longest, the question for a candidate word becomes very clean: can I break this word into pieces that are already in the dictionary? That is just the classic word-break DP on prefixes. If the answer is yes, then the word is definitely built from smaller words, because we only let earlier shorter entries participate.

The `a / aa / aaa / ...` case is a nice sanity check for that idea because every new word immediately becomes material for the next one. The `catsdogcats` and `ratcatdogcat` style cases are the other side of it: the DP is not looking for one specific split, it just needs any full segmentation to exist.

I also had to fix one fixture bug while building this. The last case accidentally declared the wrong word count, which made the real implementations silently miss `basketball`. After correcting that input, the full matrix lined up with the expected output.

PR #21099, opened 2026-04-23
Add LeetCode 479: Largest Palindrome Product

tamnd:
## Summary
This adds the `0479-largest-palindrome-product` fixtures and the language matrix entries for LeetCode.

The solution works backward from the largest possible answer by generating even-length palindromes from a descending left half. For each palindrome candidate, it checks whether the number can be factored into two `n`-digit numbers.

That is the key shift in perspective: instead of multiplying every pair of `n`-digit numbers and asking whether the product is a palindrome, we enumerate palindromes directly in descending order and stop at the first one with a valid factorization. Since the problem only asks for the answer modulo `1337`, the implementation returns `palindrome % 1337` once the factor pair is found.

## Included
- fixtures for `0479`
- explanation markdown in `tests/leetcode/x/python/0479.md`

## Test coverage
The fixtures cover all valid LeetCode inputs `n = 1..8`, which exercise the full known result sequence:
- `1 -> 9`
- `2 -> 987`
- `3 -> 123`
- `4 -> 597`
- `5 -> 677`
- `6 -> 1218`
- `7 -> 877`
- `8 -> 475`

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0479 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0479$'
```


tamnd:
This one is a good example of how much easier a search gets once you generate the right objects first.

If you start from all pairs of `n`-digit numbers, there is just too much multiplication and palindrome checking. The cleaner direction is to start from the largest palindrome shape we could hope for and walk downward. For this problem, the winning answers for `n > 1` are even-length palindromes, so taking a left half and mirroring it cuts the search space down a lot.

After that, the only question is whether the palindrome can be split into two `n`-digit factors. The nice part is that we can stop immediately on the first success because we are already moving from largest to smallest candidate.

I also like that the fixture covers the entire `1..8` range here. This problem has that funny modulo-`1337` output contract, so seeing the full sequence together makes it easier to trust that the implementation is really following the intended search instead of leaning on a hidden table.

PR #21100, opened 2026-04-23
Add LeetCode 480: Sliding Window Median

tamnd:
## Summary
This adds the `0480-sliding-window-median` fixtures and the language matrix entries for LeetCode.

The core solution uses two heaps:
- a max-heap for the lower half of the window
- a min-heap for the upper half

The tricky part is deletions, because a sliding window needs to remove the outgoing element each step and heaps do not support arbitrary deletes efficiently. The implementation handles that with lazy deletion: outgoing values are counted in a map, and stale heap tops are removed only when they reach the top.

After each add/remove step, the heaps are rebalanced so the lower half either matches the upper half in size or has one extra element. The median then comes directly from the heap tops.

To keep the fixture outputs exact and stable, medians are formatted from doubled integers, which lets integer medians stay as integers and midpoint cases print as `.5` without floating-point noise.

## Included
- fixtures for `0480`
- explanation markdown in `tests/leetcode/x/python/0480.md`

## Test coverage
The cases cover:
- the standard LeetCode example
- even window sizes with fractional medians
- repeated values and duplicate-heavy windows
- all-equal windows
- negative values with half-step medians
- a larger odd-window mixed case

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0480 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0480$'
```


tamnd:
This problem is one of those cases where the median itself is easy and the window maintenance is the real work.

Two heaps are the natural fit for tracking the lower and upper halves, but the annoying part is that the sliding window needs to eject one specific old value every step. A heap is happy to remove the top element, not some arbitrary thing from the middle. That is why the lazy-deletion map matters so much here: instead of fighting the heap, we mark outgoing values and only actually discard them when they float to the top.

Once that click happens, the rest is mostly bookkeeping. Rebalance after every add/remove, keep the lower heap allowed to hold one extra value, and the median comes straight from the heap tops. I also formatted from doubled integers rather than raw floats because this problem naturally produces a lot of `.5` answers, and I wanted the fixtures to stay exact instead of depending on language-specific float printing.

PR #21101, opened 2026-04-23
Add LeetCode 483: Smallest Good Base

tamnd:
## Summary
This adds the `0483-smallest-good-base` fixtures and the language matrix entries for LeetCode.

The key observation is that if `n` can be written as all ones in base `k`, then:

`n = 1 + k + k^2 + ... + k^m`

For a fixed number of digits, that is a geometric series. The smallest good base corresponds to the longest possible all-ones representation, so the search runs over possible lengths from large to small.

For each candidate length, the implementation binary-searches the base and evaluates the geometric sum with overflow guards. The first exact match is the answer. If no longer representation exists, the fallback is `n - 1`, which always represents `n` as `11`.

## Included
- fixtures for `0483`
- explanation markdown in `tests/leetcode/x/python/0483.md`

## Test coverage
The cases cover:
- the standard examples `13 -> 3` and `4681 -> 8`
- a very large input `10^18`, which falls back to `n - 1`
- small values that collapse to base `2`
- a mid-sized exact power-sum case `21 -> 4`
- a larger all-ones-in-base-2 case `1099511627775 -> 2`

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0483 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0483$'
```


tamnd:
This one is a neat inversion problem. The wording talks about finding a base, but the real handle is the length of the all-ones representation.

If a number is `111...111` in base `k`, then in decimal it is a geometric sum. The smaller the base, the longer that representation has to be, so the search naturally goes from the largest possible digit count downward. As soon as one length works, we are done, because any later hit would use a larger base.

The other detail that matters here is being careful with overflow while evaluating the series. For the big inputs, you really do not want to build powers naively and hope the type can absorb them. I used guarded multiplication/addition checks so the comparison against `n` stays trustworthy.

I like the `10^18` case in this batch because it highlights the fallback behavior clearly: if no longer all-ones form exists, the answer is just `n - 1`, meaning the number is `11` in that base.

PR #21102, opened 2026-04-23
Add LeetCode 488: Zuma Game

tamnd:
## Summary
This adds the `0488-zuma-game` fixtures and the language matrix entries for LeetCode.

The solution uses memoized depth-first search over two pieces of state:
- the current board after all automatic collapses
- the remaining counts of balls in hand

At each step, the search looks at one maximal run of a single color and tries the only insertion count that is ever useful there: just enough balls to make that run reach length three. After removing that run, the board is collapsed repeatedly to account for chain reactions, and the search continues from the reduced state.

That keeps the branching under control, and memoization turns repeated subproblems into cache hits.

## Included
- fixtures for `0488`
- explanation markdown in `tests/leetcode/x/python/0488.md`

## Test coverage
The cases cover:
- the standard impossible example `WRRBBW / RB`
- the standard solvable examples with chain reactions
- a single-ball board with a large matching hand
- a denser mixed-color case
- several dead-end states where the hand still cannot clear the board

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0488 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0488$'
```


tamnd:
This one is a search problem, but the part that really matters is pruning the search tree before it explodes.

The useful constraint is that there is almost never any reason to insert more balls than needed to make one existing run hit length three. If a run has length one, the only meaningful move there is to spend two of that color; if it has length two, spend one. Anything larger is just wasting hand inventory without creating a new effect. That trims the branching factor a lot.

After that, the board-collapse function does most of the interesting work because one deletion can trigger a whole chain reaction. Memoizing on the reduced board plus remaining hand counts turns out to be enough to make the recursion practical.

I also like having both kinds of examples in the fixture here: cases that look close but are still impossible, and cases where one well-placed insertion causes the board to unwind much more than it first seems.

PR #21103, opened 2026-04-24
Add LeetCode 489: Robot Room Cleaner

tamnd:
## Summary
This adds the `0489-robot-room-cleaner` fixtures and the language matrix entries for LeetCode.

Because the original LeetCode problem is interactive, this repo adapts it into a deterministic simulation. Each fixture provides:
- a room grid of open and blocked cells
- a robot start position

The implementation simulates the hidden `Robot` API and runs the standard DFS/backtracking cleaner on top of it.

The core algorithm keeps its own visited set in relative coordinates. From each reachable cell it:
- cleans the current cell
- tries all four directions
- moves into an unvisited open neighbor when possible
- explores recursively
- executes the fixed “go back” sequence so the robot returns to the previous cell facing the original direction

The result reported by the fixtures is the number of reachable cells that end up cleaned.

## Included
- fixtures for `0489`
- explanation markdown in `tests/leetcode/x/python/0489.md`

## Test coverage
The cases cover:
- a fully open room
- a cross-shaped reachable region
- denser rooms with blocked pockets and narrow passages
- a case where the start is isolated to a single reachable cell
- irregular maps that exercise turning and backtracking order

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0489 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0489$'
```


tamnd:
This problem is easy to describe and surprisingly easy to get subtly wrong in code.

The main thing the recursion has to preserve is not just position, but orientation too. After you move into a branch, explore it, and want to come back, the robot has to return to the previous cell facing the same direction it had before that branch started. That is why the fixed “turn twice, move, turn twice” backtrack step matters so much. If the heading drifts, the rest of the DFS quietly stops matching the intended geometry.

For the repo fixtures I simulated the hidden robot API directly instead of flattening the problem into a plain grid DFS. That keeps the implementation closer to what the original LeetCode solution is actually reasoning about. The interesting cases here are the ones with blocked pockets and skinny corridors, because they force the robot to rely on the backtracking discipline instead of just wandering through an open room.

PR #21104, opened 2026-04-24
Add LeetCode 493: Reverse Pairs

tamnd:
## Summary
This adds the `0493-reverse-pairs` fixtures and the language matrix entries for LeetCode.

The solution uses merge sort with counting during the merge step. After recursively sorting the left and right halves, it counts cross-half reverse pairs using a monotonic pointer in the right half:

- for each value in the sorted left half
- advance the right pointer while `left[i] > 2 * right[j]`
- add the number of right-side values seen so far

Because both halves are already sorted, that pointer only moves forward, which keeps the counting step linear per merge.

The final merge then proceeds normally. This gives `O(n log n)` time overall and avoids the quadratic scan that times out on large inputs.

## Included
- fixtures for `0493`
- explanation markdown in `tests/leetcode/x/python/0493.md`

## Test coverage
The cases cover:
- the standard LeetCode examples
- repeated negative values
- a descending negative run
- a boundary case involving `INT_MIN` and `INT_MAX`
- descending positives
- all-equal values
- a mixed unsorted case with several cross-half hits

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0493 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0493$'
```


tamnd:
This problem looks like inversion counting at first, but the `> 2 * nums[j]` condition changes just enough that you have to be careful about why the merge-sort trick still works.

The reason it survives is that once both halves are sorted, the qualifying range in the right half only moves in one direction as you scan the left half. That means the cross-half counting step is still linear, not quadratic. You count first, then do the ordinary merge.

I also wanted an explicit boundary case in the fixture because this is exactly the kind of problem where a language can accidentally overflow on `2 * nums[j]` and silently get the wrong answer. Using promoted arithmetic in the real implementations keeps that comparison honest.

The negative-value cases are nice sanity checks too. They make it obvious that reverse pairs are not just a “large positives before small positives” story; ordering around negatives can create a lot of valid pairs very quickly.

PR #21105, opened 2026-04-24
Add LeetCode 499: The Maze III

tamnd:
## Summary
Add fixtures and solutions for LeetCode 499, The Maze III, across the full language matrix.

## What changed
- add shared `0499.in` / `0499.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for how the search works
- mark `0499-the-maze-iii` complete in `tests/leetcode/TASKS.md`

## Notes
The heart of this problem is that it is not plain BFS. Each move rolls until a wall or the hole, so edges have different costs. The right model is Dijkstra over stopping positions, ordered first by total distance and then by the path string for the lexicographic tie-break.

While wiring the fixtures, one subtle bug was worth calling out explicitly: a blocked direction must not be treated as a zero-cost transition. If you allow that, you accidentally invent fake prefixes and the path order looks right until a case exposes it.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0499 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0499$'
```


tamnd:
This one was a good reminder that Maze III looks like BFS at first glance, but it really is Dijkstra once you pay attention to what a move means.

The ball does not step one cell at a time. It keeps rolling until it hits a wall or falls into the hole, so one choice can cost 1 square and the next can cost 7. That makes the graph weighted immediately. Then LeetCode adds the second wrinkle: if two routes reach the hole in the same minimum distance, we still have to return the lexicographically smaller instruction string. So the priority is really `(distance, path)` the whole way through.

While putting the fixtures together I also tripped over an easy bug shape: if a direction is blocked right away, that is not a real move. Letting it behave like a zero-cost transition quietly creates fake prefixes, and the output starts growing nonsense like an extra leading `d` even though the distance still looks plausible. Once that was fixed, the answers lined up cleanly across the matrix.

I like this problem because the solution is small once the model is right, but the model is exactly the part that wants to fool you.

PR #21106, opened 2026-04-24
Add LeetCode 502: IPO

tamnd:
## Summary
Add fixtures and solutions for LeetCode 502, IPO, across the full language matrix.

## What changed
- add shared `0502.in` / `0502.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the greedy heap approach
- mark `0502-ipo` complete in `tests/leetcode/TASKS.md`

## Notes
This problem looks like general project scheduling at first, but the useful structure is narrower: at any moment, the only decision that matters is which currently affordable project has the largest profit. Sorting by required capital and using a max-heap for affordable profits makes that choice explicit.

The fixture set covers the main edges here: unlocking future projects with an early pick, stopping immediately when nothing is affordable, and cases where `k` is larger than the number of profitable moves that actually matter.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0502 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0502$'
```


tamnd:
This one is a nice example of a greedy choice that feels suspicious until you phrase the state correctly.

At first it sounds like we might need to plan several projects ahead, because one small project can unlock a much bigger one later. But once you sort projects by required capital, the state at any step becomes simple: among everything I can already afford, which pick gives me the most capital for the next round? There is no reason to take a smaller profit over a larger profit when both are available, because the larger one only makes the future option set at least as good.

So the whole solution turns into two moving parts: a pointer through projects ordered by capital requirement, and a max-heap of profits for the projects that are currently unlocked. Each round, push everything newly affordable, pop the best profit, add it to capital, repeat up to `k` times.

I also liked building fixtures for the dead-end case here. If the heap is empty, the algorithm should just stop immediately instead of trying to force exactly `k` picks. That early stop is really part of the problem, not an implementation detail.

PR #21107, opened 2026-04-24
Add LeetCode 514: Freedom Trail

tamnd:
## Summary
Add fixtures and solutions for LeetCode 514, Freedom Trail, across the full language matrix.

## What changed
- add shared `0514.in` / `0514.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the ring-position DP
- mark `0514-freedom-trail` complete in `tests/leetcode/TASKS.md`

## Notes
The useful state here is not the whole rotated ring string. It is just which ring index currently sits at twelve o'clock after spelling a prefix of the key. Once that state is chosen, the next move cost is only the shorter circular distance to one of the matching positions for the next character.

The fixture set covers repeated letters, multiple occurrences of the same character in the ring, and the easy place to make a mistake: two different matching positions can have different downstream costs even when they look equally good locally.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0514 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0514$'
```


tamnd:
This problem is a good example of a DP that gets much simpler once you stop carrying around too much state.

My first instinct was to think in terms of the whole rotated ring, but that is unnecessary. After spelling some prefix of the key, the only thing that matters is which ring index is sitting at twelve o'clock. From there, the next character means: try every matching copy of that character in the ring, pay the shorter clockwise/counterclockwise distance to reach it, and keep the best total.

The interesting part is that matching positions are not interchangeable. Two copies of the same letter can cost the same to reach now, but one can be much better for the remaining suffix of the key. That is why the DP is over positions, not just characters.

I also had to fix two expected values in the fixture set while wiring this up. The solver was right; my hand calculation for the last two cases was not. That was actually a nice sanity check for this problem, because it is easy to underestimate how much future alignment matters after one locally cheap move.

PR #21108, opened 2026-04-24
Add LeetCode 517: Super Washing Machines

tamnd:
## Summary
Add fixtures and solutions for LeetCode 517, Super Washing Machines, across the full language matrix.

## What changed
- add shared `0517.in` / `0517.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the prefix-balance greedy scan
- mark `0517-super-washing-machines` complete in `tests/leetcode/TASKS.md`

## Notes
The key step here is to stop thinking in terms of simulating moves. What matters is the imbalance each machine contributes relative to the target average, and the cumulative flow that must cross each boundary between machines.

The answer comes from the maximum of two quantities during a left-to-right scan: the absolute running balance across a boundary and the local surplus of a machine that has to push dresses out. The fixtures cover both impossible totals and cases where the bottleneck is not the biggest pile itself but the amount of flow that has to pass through part of the line.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0517 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0517$'
```


tamnd:
This one looks like it wants an explicit move simulation, but that is exactly the trap.

The useful mental model is flow. After fixing the target average, each machine either has a surplus to send out or a deficit that has to be filled. When you walk left to right, the running prefix balance tells you how many dresses must cross the boundary after this machine. That boundary flow is unavoidable, no matter how cleverly you schedule individual moves.

Then there is a second constraint that sits on top of that: a machine with a large positive surplus can only send one dress per move to one adjacent machine, so its own local surplus can also be the bottleneck. That is why the answer is the max of `abs(prefix_balance)` and the current machine surplus.

I also had to correct two fixture answers while wiring this up. My first guesses for the last two cases were too optimistic because they ignored the amount of flow that had to propagate through the line. The reference implementation made that mistake obvious right away, which is exactly the kind of check I want from these batches.

PR #21109, opened 2026-04-24
Add LeetCode 527: Word Abbreviation

tamnd:
## Summary
Add fixtures and solutions for LeetCode 527, Word Abbreviation, across the full language matrix.

## What changed
- add shared `0527.in` / `0527.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the grouped prefix/LCP approach
- mark `0527-word-abbreviation` complete in `tests/leetcode/TASKS.md`

## Notes
The key simplification in this problem is that possible conflicts only live inside groups with the same word length and the same first and last character. Inside one of those groups, sorting the words lets us find the minimum visible prefix for each word just from its common-prefix lengths with its neighbors.

The fixture set covers the classic example, words that should stay unchanged because abbreviating does not help, and groups where some words need longer prefixes while others in the same bucket can still keep very short abbreviations.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0527 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0527$'
```


tamnd:
This problem gets much cleaner once you notice where conflicts can and cannot exist.

Two words can only ever collide as abbreviations if they have the same length, the same first character, and the same last character. That means the whole global problem breaks into independent little buckets. After that, sorting inside one bucket is enough: for each word, the only neighbors that matter are the ones right next to it lexicographically, because those are the words that can share the longest prefix with it.

So the real job is not repeatedly bumping prefix lengths until things stop colliding. It is figuring out the minimum prefix each word must expose, which is just one more than the maximum common-prefix length with its adjacent words in the sorted bucket.

I like this one because it looks like it wants a conflict-resolution loop, but there is a much more structural way to see it. Once the grouping is right, the rest is almost bookkeeping. The fixture set here tries to show that mix: some words stay unchanged because abbreviating would not help, while others in the same family need just enough extra prefix to break the tie and no more.

PR #21111, opened 2026-04-28
Add LeetCode 564: Find the Closest Palindrome

tamnd:
## Summary
Add fixtures and solutions for LeetCode 564, Find the Closest Palindrome, across the full language matrix.

## What changed
- add shared `0564.in` / `0564.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the small mirrored-candidate set
- mark `0564-find-the-closest-palindrome` complete in `tests/leetcode/TASKS.md`

## Notes
The useful observation here is that the nearest answer cannot come from an arbitrary search around `n`. It must be one of a few mirrored numbers built from the left half of `n`, plus the two length-boundary cases like `999...9` and `1000...0001`.

The fixture set leans on exactly those edge cases: single digits, powers of ten, all nines, already-palindromic inputs, and a large 18-digit example to keep the implementations honest about 64-bit handling.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0564 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0564$'
```


tamnd:
This problem looks like it might need a search around `n`, but the real solution is much tighter than that.

The nearest palindrome has to be structurally close, not numerically brute-forced. In practice that means taking the left half, mirroring it, and also checking what happens if that left half goes up or down by one. Then there are the two annoying but necessary edge cases: crossing a digit-length boundary, which is where `99...9` and `100...001` come from.

What I like about this one is that the hard part is not generating palindromes, it is trusting that you only need a handful of them. Once that observation is solid, the rest becomes careful tie-breaking: exclude the number itself, minimize absolute difference, and on ties return the smaller palindrome.

The fixture set here leans hard on the cases that usually break first attempts: single digits, powers of ten, all nines, already-palindromic inputs, and one large 18-digit value just to make sure the implementations stay honest about integer range.

PR #21112, opened 2026-04-28
Add LeetCode 568: Maximum Vacation Days

tamnd:
## Summary
Add fixtures and solutions for LeetCode 568, Maximum Vacation Days, across the full language matrix.

## What changed
- add shared `0568.in` / `0568.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the week-by-week DP
- mark `0568-maximum-vacation-days` complete in `tests/leetcode/TASKS.md`

## Notes
The useful state here is just: which city can we start this week in, and what is the best vacation total so far? That makes the natural solution a DP over weeks, where each step considers staying put or taking any available flight before collecting that week's vacation days in the destination city.

The fixture set includes the classic example, no-flight cases, diagonally favorable schedules, and a couple of small graphs where the best total comes from planning the city sequence across multiple weeks rather than grabbing the largest immediate value.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0568 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0568$'
```


tamnd:
This one is a nice reminder that not every graph-shaped problem wants graph search as the main abstraction.

You can think of the flights as a graph, but the more useful structure is time. Each week is one layer, and the only question that matters at that layer is: if I begin this week in city `c`, what is the best total vacation I could have accumulated so far? From there, the transition is straightforward: stay where you are or take any allowed flight, then collect that week's days in the destination city.

What I like here is that the DP strips away a lot of noise. We do not need to remember the whole route, only the best total attached to each reachable city for the current week. That is enough because future choices depend only on current location and remaining weeks.

The fixture set tries to cover the different moods of the problem: cases with no flights at all, cases where moving every week is clearly best, and cases where the optimal plan is less obvious because a smaller short-term choice sets up a better city for later weeks.

PR #21113, opened 2026-04-28
Add LeetCode 587: Erect the Fence

tamnd:
## Summary
Add fixtures and solutions for LeetCode 587, Erect the Fence, across the full language matrix.

## What changed
- add shared `0587.in` / `0587.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the boundary-preserving monotonic chain hull
- mark `0587-erect-the-fence` complete in `tests/leetcode/TASKS.md`

## Notes
This is the convex hull problem with one important twist: trees that lie on a straight boundary segment still belong in the answer. So the monotonic-chain hull has to pop only on strictly clockwise turns, not on collinear ones.

The fixture set covers the standard example, fully collinear input, a single-point case, and a couple of shapes where several trees sit along the same fence edge and need to stay in the output.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0587 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0587$'
```


tamnd:
The interesting part of this problem is that it looks like a standard convex hull until the output requirement changes the usual implementation detail.

If we were only asked for the corner vertices, the monotonic-chain hull would pop on non-left turns and be done. But here every tree lying on the fence perimeter has to stay in the answer, including trees that sit on a straight edge between two extreme corners. That means collinear points are not noise, they are part of the result.

So the whole solution really turns on one small but important choice: pop only on strictly clockwise turns. Once I keep that invariant in mind, the rest is the usual lower-hull plus upper-hull construction, followed by deduping the combined set.

I like this one because it is a geometry problem where the bug is rarely in the geometry formula itself. The bug is usually in the interpretation of what counts as being on the hull. The fixture set here is mostly about that, especially the cases with several points lined up along the same boundary edge.

PR #21114, opened 2026-04-28
Add LeetCode 600: Non-negative Integers without Consecutive Ones

tamnd:
## Summary
Add fixtures and solutions for LeetCode 600, Non-negative Integers without Consecutive Ones, across the full language matrix.

## What changed
- add shared `0600.in` / `0600.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the bit-DP / Fibonacci counting approach
- mark `0600-non-negative-integers-without-consecutive-ones` complete in `tests/leetcode/TASKS.md`

## Notes
The useful counting fact here is that the number of valid binary strings of a fixed length follows the Fibonacci recurrence. Once that table is built, we can scan the bits of `n` from high to low, adding the counts for smaller prefixes and stopping as soon as `n` itself contains `11`.

The fixture set covers the examples, values around powers of two, all-ones boundaries like `1023`, and a large `10^9` case to make sure the implementations stay honest on the upper end of the input range.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0600 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0600$'
```


tamnd:
What I like about this problem is that it looks like a binary-digit DP, but underneath it there is a very small counting pattern.

If you ask only for the number of binary strings of length `k` with no consecutive ones, you immediately get a Fibonacci recurrence: strings starting with `0` can use any valid suffix of length `k-1`, and strings starting with `10` can use any valid suffix of length `k-2`. That table is the whole engine.

Then the scan over `n` is basically bookkeeping. Walk bits from left to right, and whenever you see a `1`, add the count for the branch where this bit becomes `0` and the rest is any valid suffix. If you ever encounter two consecutive `1`s in `n`, you can stop right there because every continuation of that exact prefix is already invalid.

I like this one because the final code is short, but only after the counting model is clean. The fixture set here leans into the places where that model matters most: powers of two, all-ones boundaries, and numbers just above those boundaries where the count bumps by exactly one.

PR #21115, opened 2026-04-28
Add LeetCode 629: K Inverse Pairs Array

tamnd:
## Summary
Add fixtures and solutions for LeetCode 629, K Inverse Pairs Array, across the full language matrix.

## What changed
- add shared `0629.in` / `0629.out` fixtures for all supported LeetCode language runners
- add the Python algorithm note for the sliding-window DP
- mark `0629-k-inverse-pairs-array` complete in `tests/leetcode/TASKS.md`

## Notes
The direct recurrence says that when we insert `n`, it can contribute anywhere from `0` to `n-1` new inverse pairs depending on position. The naive DP therefore sums a range of previous states for every `(n, k)`.

The useful optimization is that those ranges slide by one as `k` increases, so we can maintain the sum with a running window instead of recomputing it each time. That turns the solution from O(nk²) into O(nk), which is the difference between timing out and finishing comfortably under the constraints.

## Verification
```sh
PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0629 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0629$'
```


tamnd:
This is one of those DP problems where the recurrence is easy to write down and the real work is making it fast enough.

If we build permutations of `1..n` by inserting `n` into a permutation of `1..n-1`, the new element can create anywhere from `0` to `n-1` inverse pairs depending on where it lands. That gives the natural recurrence immediately, but it also makes the naive transition a sum over a whole range for every state.

The nice part is that those ranges overlap almost completely as `k` increases. So instead of recomputing the sum each time, we keep a sliding window over the previous DP row: add the new right endpoint, subtract the value that just fell off the left. Same recurrence, much cheaper execution.

I like this problem because it is a clean example of how a good optimization often comes from staring at the transition shape rather than inventing a new algorithm. The fixture set here includes a couple of larger cases mainly to force that distinction: the recurrence is not enough by itself, the running-window implementation is the point.

PR #21116, opened 2026-04-28
Add LeetCode 630: Course Schedule III

tamnd:
## Summary
- add LeetCode 630 fixtures, explanation, and implementations across the language matrix
- solve the problem with the standard deadline-ordered greedy plus max-heap replacement strategy

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0630 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0630$'`


tamnd:
This one is a nice greedy problem because the obvious first instinct is incomplete. Sorting by deadline is necessary, but it is not enough to just keep taking courses while they fit.

The real move is: take each course in deadline order, pretend we keep it, and only when the total time spills past the current deadline do we undo the most expensive choice so far. That means removing the longest-duration course seen up to that point, not necessarily the one we just added.

What makes that work is that for any fixed number of courses, the best schedule to preserve future flexibility is the one with the smallest total occupied time. The max-heap gives that adjustment in one step every time we go over budget.

I like this problem because the final algorithm is short, but the key invariant is doing all the work: after each deadline, we keep the largest possible number of courses among everything processed so far, and among those schedules we keep the least total time.

PR #21117, opened 2026-04-28
Add LeetCode 631: Design Excel Sum Formula

tamnd:
## Summary
- add LeetCode 631 fixtures, explanation, and implementations across the language matrix
- model cells as either plain values or formula references so later writes flow through dependent sums

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0631 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0631$'`


tamnd:
This problem is really about what a cell means after it stops being just a number.

A normal `set` is easy: the cell becomes a plain value and whatever formula used to live there is gone. The interesting part is `sum`, because that cell is no longer storing an answer so much as a recipe. It remembers which cells and ranges contribute, including duplicates, and its value has to be interpreted against the current sheet every time we ask for it.

That is why I kept the cell representation split in two forms: either a raw integer or a counted map of dependencies. Once you do that, the behavior after later updates becomes natural. If `C3` was defined from `A1` and `A1:B2`, and then `B2` changes, there is nothing special to propagate by hand in these tests. `get(C3)` just re-evaluates the recipe and sees the new sheet state.

The cases here try to hit the parts that are easy to get subtly wrong: overwriting a formula cell with `set`, counting the same reference more than once, expanding ranges correctly, and chaining one formula through another formula cell.

PR #21118, opened 2026-04-28
Add LeetCode 632: Smallest Range Covering Elements from K Lists

tamnd:
## Summary
- add LeetCode 632 fixtures, explanation, and implementations across the language matrix
- solve the problem with the standard min-heap over current list heads while tracking the current maximum

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0632 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0632$'`


tamnd:
This one looks like a sliding-window problem at first, but the heap view is the cleaner way to think about it.

At any moment we keep exactly one chosen value from each list. Those chosen values already define a valid covering range: from the smallest chosen value to the largest chosen value. So the question becomes how to move from one valid cover to the next without missing the optimum.

The only pointer that can possibly improve the range is the one sitting on the current minimum. Advancing any other list would keep the same left edge and only risk pushing the right edge farther out. That is the whole reason the min-heap works here: pop the current minimum, try the next value from that same list, update the running maximum, and compare the new range.

I also like that the tie-break is part of the actual problem, not a detail to bolt on later. When two ranges have the same width, we need the one with the smaller left endpoint, so the comparison has to encode that from the start instead of treating every equal-width interval as interchangeable.

PR #21119, opened 2026-04-28
Add LeetCode 639: Decode Ways II

tamnd:
## Summary
- add LeetCode 639 fixtures, explanation, and implementations across the language matrix
- solve the wildcard decoding count with constant-space dynamic programming and explicit one-character and two-character contribution counts

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0639 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0639$'`


tamnd:
This problem is one of those DP questions where the recurrence is short, but only after you do the casework honestly.

A plain digit-only version is already about whether the last step used one character or two. The `*` wildcard just turns those two transitions into weighted counts instead of yes-or-no choices. So I found it cleaner to separate the reasoning into two helper questions:

1. how many single-letter decodes can this character contribute by itself?
2. how many two-letter decodes can this pair contribute together?

Once those counts are correct, the DP is almost mechanical: current ways = single-count times previous ways + pair-count times ways from two positions back.

The spots that matter most are exactly the annoying ones: `0` never stands alone, `*0` has only the `10` and `20` interpretations, `**` contributes 15 valid two-digit combinations, and longer strings only stay manageable if you keep everything modulo `1e9+7` from the start. I like this problem because it looks like a string puzzle, but the real challenge is just being disciplined about the combinatorics.

PR #21120, opened 2026-04-28
Add LeetCode 642: Design Search Autocomplete System

tamnd:
## Summary
- add LeetCode 642 fixtures, explanation, and implementations across the language matrix
- model the autocomplete stream with sentence frequencies, a live prefix buffer, and top-3 ranking by hotness then lexicographic order

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0642 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0642$'`


tamnd:
What I like about this problem is that the ranking rule is more important than the data structure choice in a small test harness.

Every keystroke is really asking for the same thing: among all sentences that start with the current prefix, return the three best by hotness, and if hotness ties, break ties lexicographically. The `#` character is the second half of the design, because it turns the buffered prefix into a new sentence event and mutates the ranking for future queries.

So the key state is just two pieces: the sentence frequency table and the current prefix buffer. Once those are right, each `input` call becomes very direct. Extend the prefix, collect matches, sort by `(-count, sentence)`, and trim to three. On `#`, push the accumulated sentence back into the table and reset the buffer.

The cases here are mainly about behavior rather than complexity: space characters inside sentences, ties that must be resolved alphabetically, newly committed phrases showing up in later suggestions, and existing phrases getting promoted when they are typed again.

PR #21121, opened 2026-04-28
Add LeetCode 644: Maximum Average Subarray II

tamnd:
## Summary
- add LeetCode 644 fixtures, explanation, and implementations across the language matrix
- solve the problem with binary search on the average plus a linear prefix-sum feasibility check for subarrays of length at least k

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0644 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0644$'`


tamnd:
This is one of those problems where the key idea is to stop chasing the subarray directly and instead ask a yes-or-no question about a candidate average.

If we guess an average `mid`, then subtracting `mid` from every element changes the problem into: does there exist a subarray of length at least `k` with non-negative sum? That reformulation is the real heart of the solution.

From there, the prefix-sum check is nice and tight. As we scan left to right, we keep the transformed prefix sum for the current end, and we also keep the minimum transformed prefix that is far enough behind to leave at least `k` elements in between. The moment `current_prefix - minimum_previous_prefix >= 0`, we know some valid subarray reaches average at least `mid`.

I like this problem because the binary search is not the interesting part by itself. The interesting part is proving that the feasibility test is linear and complete. Once that clicks, the rest is just numerical refinement, which is why fixed decimal output works well for the harness here.

PR #21122, opened 2026-04-28
Add LeetCode 656: Coin Path

tamnd:
## Summary
- add LeetCode 656 fixtures, explanation, and implementations across the language matrix
- solve the problem with backward dynamic programming that minimizes total cost and breaks ties by lexicographically smaller paths

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0656 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0656$'`


tamnd:
This problem looks like a shortest-path variant, but the lexicographic tie-break is what makes the DP shape matter.

Working backward is the clean way to express it. For each index, we ask: among all reachable next positions within `maxJump`, which one gives the minimum total cost to the end? If several give the same cost, we keep the lexicographically smaller full path.

That second rule is the part that is easy to under-appreciate. Two next jumps can have the same remaining cost, but the required answer is not arbitrary. It has to prefer the earlier path in dictionary order of indices, which is why storing the chosen suffix path itself keeps the comparison honest in the harness.

The cases here try to hit exactly those edge behaviors: blocked cells with `-1`, completely unreachable endings, zero-cost paths, and ties where the answer is determined by path order rather than cost alone.

PR #21123, opened 2026-04-28
Add LeetCode 660: Remove 9

tamnd:
## Summary
- add LeetCode 660 fixtures, explanation, and implementations across the language matrix
- solve the problem by interpreting the answer as the base-9 representation of n written with decimal digits

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0660 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0660$'`


tamnd:
This problem is short once you see the numbering system hiding inside it.

If you list the positive integers while skipping every number that contains the digit `9`, you are effectively counting in base 9. The first few terms make that visible: after `8`, the next valid number is `10`, then `11`, and so on, exactly like base-9 carries but written with ordinary decimal characters.

So the solution is not really about searching or filtering numbers at all. It is just: convert `n` to base 9, then read that digit string as the answer. That is why cases like `9 -> 10`, `18 -> 20`, and `81 -> 100` matter so much in the fixtures: they are the carry points where the hidden base changes shape.

I like this one because it looks like a digit-DP problem at first glance, but the clean solution is a much simpler observation about how the target sequence is generated.

PR #21124, opened 2026-04-28
Add LeetCode 675: Cut Off Trees for Golf Event

tamnd:
## Summary
- add LeetCode 675 fixtures, explanation, and implementations across the language matrix
- solve the problem by sorting trees by height and summing BFS distances between consecutive targets

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0675 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0675$'`


tamnd:
What makes this problem feel bigger than it is is the map, but the order is already fixed for us.

Once the trees have to be cut from shortest to tallest, there is no global route optimization left to invent. The job becomes a sequence of local movement problems: from where I am now, what is the shortest walk to the next required tree?

That is why the clean structure is just two layers. First, collect and sort all tree cells by height. Second, for each consecutive pair of positions, run a BFS on the current walkable grid and add that distance to the total. If any one of those BFS runs fails, the whole answer is `-1` immediately.

I like the third sample shape in particular because it highlights a subtle point: if the first cell already contains the smallest tree, cutting it costs zero movement. The difficulty is not in the cut operation itself, it is entirely in the movement between mandatory targets.

PR #21125, opened 2026-04-28
Add LeetCode 679: 24 Game

tamnd:
## Summary
- add LeetCode 679 fixtures, explanation, and implementations across the language matrix
- solve the problem with recursive search over pairwise arithmetic combinations using floating-point states and an epsilon check around 24

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0679 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0679$'`


tamnd:
This problem is a good reminder that the search space sounds bigger than it really is.

At any point we only care about the current multiset of values still available to combine. So the natural move is: pick two numbers, replace them with every result obtainable from `+`, `-`, `*`, and valid `/`, and recurse on the shorter list. Starting from four cards, that tree stays small enough to brute force comfortably.

The two practical details that matter are symmetry and floating-point tolerance. Addition and multiplication are symmetric, but subtraction and division are not, so both operand orders have to be explored there. And because division is real-valued, the base case should not demand exact equality with `24`; it needs a small epsilon window around it.

The nice thing about the fixtures here is that they show different reasons for `true`: some are obvious multiplication-based wins, while others only work once division reshapes the intermediate values. The `false` cases are useful because they make sure the search is actually exhaustive instead of just pattern-matching common formulas.

PR #21126, opened 2026-04-28
Add LeetCode 683: K Empty Slots

tamnd:
## Summary
- add LeetCode 683 fixtures, explanation, and implementations across the language matrix
- solve the problem with the days-array transformation plus a linear sliding-window scan over candidate endpoints

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0683 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0683$'`


tamnd:
The nice trick in this problem is to stop thinking in bloom order and start thinking in bloom day per position.

Once `days[i]` means “when does slot `i` bloom?”, a valid answer for distance `k` is just a window whose two ends bloom earlier than every flower strictly inside the window. If that happens, then on the later of those two endpoint days, the endpoints are on and the inside is still off.

That turns the problem into a clean scan over windows of width `k + 2`. When an interior slot blooms too early, it kills the current window and becomes the natural new left boundary, because any window spanning past it would already be invalid. That is why the scan stays linear instead of rechecking everything from scratch.

I like this problem because the original statement sounds like it wants a dynamic set or tree structure, but the days-array view collapses it into a very direct sliding-window condition.

PR #21127, opened 2026-04-28
Add LeetCode 685: Redundant Connection II

tamnd:
## Summary
- add LeetCode 685 fixtures, explanation, and implementations across the language matrix
- solve the problem by separating the directed graph into the two-parent case, the pure cycle case, and the combined case, using union-find to decide which edge to remove

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0685 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0685$'`


tamnd:
This problem is tricky mostly because there are really three different graph failures hiding under one statement.

A directed rooted tree can break in two ways after one extra edge is added: some node can end up with two parents, or a cycle can appear. The annoying case is when both happen together, because then you cannot decide from local degree information alone which edge is actually redundant.

The clean approach is to separate those concerns. First scan for a node with two incoming edges and remember both competing parent edges. Then run union-find across the graph, temporarily skipping the later competing edge if it exists. If a cycle still shows up, that means the earlier parent edge is the real problem. If no cycle shows up, then the later competing edge is the one to remove. And if there was never a two-parent conflict in the first place, the edge that closes the cycle is the answer.

I like this problem because the union-find itself is straightforward; the real work is being precise about which structural case you are actually in.

PR #21129, opened 2026-04-28
Add LeetCode 691: Stickers to Spell Word

tamnd:
## Summary
- add LeetCode 691 fixtures, explanation, and implementations across the language matrix
- solve the problem with memoized DFS over target bitmasks, using each sticker to advance the covered-state as far as possible

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0691 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0691$'`


tamnd:
What I like in this problem is that the target is small, so the real state is not “which stickers have I used” but simply “which target positions are already covered”. Once you look at it that way, the search space becomes much cleaner.

Each sticker is just a way to push that covered-state forward. For a given mask of finished positions, I take one sticker, spend its letters greedily on the still-missing characters, and get the next mask. Memoizing that minimum over all masks is the whole solution.

The small pruning detail that matters is only trying stickers that can cover the first uncovered target character. If a sticker cannot help with the next missing letter, it cannot be the first move of an optimal continuation from that state either, so skipping it keeps the recursion from wandering around useless branches.

PR #21130, opened 2026-04-28
Add LeetCode 699: Falling Squares

tamnd:
## Summary
- add LeetCode 699 fixtures, explanation, and implementations across the language matrix
- solve the problem by scanning earlier overlapping squares to compute each new landing height and track the running maximum skyline

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0699 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0699$'`


tamnd:
This one looks geometric at first, but the actual question is simpler: when a new square falls, which earlier tops can it possibly land on?

The only squares that matter are the ones whose horizontal intervals overlap the new interval. Among those, the square lands on the highest top, and then we add its own side length. So every drop is really just “find the tallest overlapping support”.

What makes the output a little different from the per-square heights is that the problem wants the skyline after each drop, not just the latest square’s height. That is why the running maximum is tracked separately. A square can land low while the global skyline stays unchanged, or it can stack on top of an older tower and push the whole profile upward.

PR #21131, opened 2026-04-28
Add LeetCode 710: Random Pick with Blacklist

tamnd:
## Summary
- add LeetCode 710 fixtures, explanation, and implementations across the language matrix
- solve the problem with blacklist remapping from the valid prefix into unused tail values, using a deterministic pick stream for stable cross-language tests

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0710 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0710$'`


tamnd:
The clean idea here is that we do not need to sample from all `n` values and reject blacklisted ones. We only need to sample from the count of valid answers.

If `m = n - blacklist.size()`, then every pick can be thought of as choosing a number in `[0, m)`. Most of those are already valid. The only awkward ones are the blacklisted numbers that also lie in that prefix. Those are the ones we remap.

So the constructor is really just bookkeeping: find unused values in the tail `[m, n)`, and pair them with the blocked values in `[0, m)`. After that, `pick()` is tiny. Draw one number in the compact valid range, and if it is one of the blocked prefix values, jump through the remap table. Otherwise keep it as-is. That is what turns the problem from rejection sampling into O(1) sampling after O(b) setup.

PR #21132, opened 2026-04-28
Add LeetCode 711: Number of Distinct Islands II

tamnd:
## Summary
- add LeetCode 711 fixtures, explanation, and implementations across the language matrix
- solve the problem by canonicalizing each island across all eight rotations and reflections before counting unique signatures

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0711 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0711$'`


tamnd:
The heart of this problem is deciding what it means for two islands to be “the same” once rotation and reflection are allowed.

If you only store the raw DFS walk or the raw coordinates, that notion is hidden. The cleaner view is to say that each island has eight possible geometric presentations: four rotations, and then those same four after mirroring. All we need is one canonical representative among those eight.

So after collecting an island’s cells, I transform the coordinates in all eight ways, shift each transformed set so it starts at the origin, sort the points, and take the lexicographically smallest normalized list. Two islands are equivalent exactly when that canonical form matches. Once that normalization is in place, the rest of the problem is just flood-fill plus a set of signatures.

PR #21133, opened 2026-04-28
Add LeetCode 715: Range Module

tamnd:
## Summary
- add LeetCode 715 fixtures, explanation, and implementations across the language matrix
- solve the design problem with a sorted disjoint interval list supporting merge, containment query, and split-on-remove

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0715 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0715$'`


tamnd:
What makes this design problem manageable is the decision to keep the internal state as sorted disjoint intervals, not raw individual numbers and not arbitrary overlapping segments.

With that representation, `addRange` is just a merge with the neighboring intervals it touches, `queryRange` is a containment check against one candidate interval, and `removeRange` is the mirror image of add: every overlapping interval either disappears, gets trimmed on one side, or splits into two smaller disjoint pieces.

So the whole problem is really about maintaining that invariant. Once the interval list stays sorted and disjoint after every operation, each method becomes much simpler than it first looks.

PR #21134, opened 2026-04-28
Add LeetCode 716: Max Stack

tamnd:
## Summary
- add LeetCode 716 fixtures, explanation, and implementations across the language matrix
- solve the design problem with a direct stack plus temporary buffer so `popMax` removes the top-most occurrence of the maximum

## Testing
- `PATH="/opt/homebrew/opt/openjdk@17/bin:$PATH" MOCHI_ROSETTA_ONLY=0716 go test ./tests/leetcode -tags slow -run 'Test(C|Cpp|Csharp|Clojure|Dart|Elixir|Erlang|Go|Haskell|Java|Kotlin|Lean|Lua|Mochi|Ocaml|Pascal|PHP|Python|Ruby|Rust|Scala|Swift|TypeScript|Zig)Solutions/0716$'`


tamnd:
The detail that makes this problem slightly different from an ordinary stack-with-maximum is the `popMax` rule: when the maximum appears multiple times, we have to remove the one closest to the top.

That means it is not enough to know what the maximum value is. We also need to preserve stack order around that particular occurrence. The simple way to do it is to peel values off the top into a buffer until we hit the first maximum, remove that one, and then restore the buffered values.

So the real invariant is still just the stack order. `peekMax` tells us which value we are hunting for, and the temporary buffer lets `popMax` delete exactly the right copy without disturbing the relative order of everything above it.

PR #21221, opened 2026-05-11
mep-0001, mep-0002, mep-0003: tighten prose and harden AST contract

### LINKED ISSUE #21220: mep-0003: deep rewrite; soundness gaps; AST invariant validator
MEP-3 currently reads as a catalogue of Go struct definitions with a thin Motivation. It does not explain why the AST has the shape it has, what soundness obligations the shape implies, or which obligations are not yet enforced. After MEP-1 and MEP-2 were rewritten in the PEP 617 style with explicit conformance obligations, MEP-3 is the weakest of the three foundational documents.

Rewrite MEP-3 as a design proposal. Quality target: PEP. High-level rationale plus deep technical detail; no repeated info; no inline catalogue that MEP-2 already gives in grammar form.

Scope:

1. **Why these structural choices.** Tagged-union-by-nullable-field is a choice with tradeoffs; document them. The parser-AST coupling is a choice with tradeoffs; document them. The flat BinaryExpr with post-parse precedence is a choice with tradeoffs; document them.

2. **Structural soundness obligations.** What invariants must hold on a well-formed AST. Each obligation must be testable: a third party can write a validator that asserts it.

3. **Known soundness gaps.** Enumerate the concrete cases where the AST today admits shapes that have no meaning, and propose a fix for each. Audit findings to incorporate:
   - Tagged unions not asserted exactly-one-arm after parse.
   - MatchCase.Pattern is *Expr; only literal / identifier / call shapes are legal patterns. No node-level enforcement.
   - TypeDecl.Members vs Variants vs Alias mutual exclusion is documented but not asserted.
   - Literal.Null is a bool while siblings are pointers; structurally Null:true and Int:&x can coexist.
   - LoadExpr/SaveExpr Path *string makes "no path" vs "empty path" representable but semantically indistinguishable at runtime.
   - Several nodes lack Pos (TypeRef, Param, PostfixOp, BinaryExpr, PostfixExpr, AgentBlock, type-level nodes), so error reporting points at the wrong span.

4. **Code fixes shipped with this proposal.**
   - Add a tagged-union invariant validator that runs after parser.Parse and asserts each tagged union has exactly one arm set.
   - Reject load \"\" as T and save x to \"\" at normalize-time so empty path stops being a synonym for absent path.

Out of scope for this PR; tracked as open gaps:
- Pattern AST as a sealed type.
- Type attribution on AST nodes.
- Adding Pos to every node.
- Balancing BinaryExpr into a precedence tree at parse time.

tamnd:
## Summary

- Rewrite MEP-3 as a proposal-style document. The previous catalogue is replaced with a Specification of the AST data shape, a Design Principles section explaining each encoding choice (tagged-union-by-nullable-field, flat binary spine, AST owned by parser, nullable scalar for tri-state path semantics), a Soundness section listing the invariants the parser now enforces, and the soundness gaps that remain.
- Reject `load \"\" as T` and `save x to \"\"` at parse time with a new diagnostic (\`P062\`). The empty-string path was structurally indistinguishable from the absent form and surfaced as unrelated I/O errors downstream.
- Add \`parser.assertProgramInvariants\`: a deterministic post-normalisation pass that asserts the exactly-one-arm property on every tagged-union node in the AST (Statement, Primary, TypeRef, TypeDecl, PostfixOp, Literal, FunExpr body) and the SingleVariant-was-lifted property on TypeDecl. Violations surface as \`P070\` with a position pointing at the offending node.
- Wire the validator into \`parser.Parse\` and \`parser.ParseString\` so every consumer receives a tree the contract holds on.
- Add fixtures for the two new \`P062\` cases and internal-package tests covering the validator's exactly-one-arm property across Statement, TypeRef, and TypeDecl.
- Sweep MEP 1/2/3 for AI-tell prose patterns (\"X is not Y, it is Z\", em-dashes, telegraphic three-beat sentences) and drop standalone PEP citations.
- Make every \"MEP N\" cross-reference a clickable link.

Closes #21220.

## Test plan

- [x] \`go test ./parser/... ./types/... ./ast/... ./interpreter/...\` passes.
- [x] New \`tests/parser/errors/load_empty_path.{mochi,err}\` and \`tests/parser/errors/save_empty_path.{mochi,err}\` produce \`P062\` at the correct position.
- [x] New internal tests in \`parser/invariants_test.go\` exercise: Statement multi-arm, Statement no-arm, TypeDecl with un-lifted SingleVariant, TypeRef multi-arm, and a happy-path corpus through \`ParseString\`.
- [x] The previous \`tests/types/errors/let_missing_type_or_value\` continues to surface its \`T000\` diagnostic; the parser-level invariant was scoped to true tagged-union violations only.


tamnd:
Cycle 1 done: doc-comment attribution rule (gap 7) now has a written contract in mep-0003 under §"Doc-comment attribution", and gap 7 is dropped from the remaining-gaps list. Also fixed an old MDX hazard in mep-0005 where `{int, float, int64}` was being parsed as a JSX expression by Docusaurus and breaking website build. Both CI checks green.

Moving on to cycle 2 next: add Pos fields to LogicCond/PostfixOp/AgentBlock and extend the AST validator to cover LogicCond.

tamnd:
Cycle 2 done: three more gaps closed.

- LogicCond, PostfixOp, AgentBlock all carry their own Pos now. The grammar tag is just a placeholder, participle fills it in. A diagnostic on a rule body or "the third subscript in a chain" can point at the right node instead of leaking down to the first child.
- The validator walks rule bodies and agent bodies and asserts the exactly-one-of rule on LogicCond and AgentBlock. Two new tests in invariants_test.go hand-build a bad node of each shape and check we reject it with P070. The valid corpus picked up a `rule` and `fact` program so we exercise the happy path too.
- mep-0003 §"Specification" got the new Pos fields in the type sketches and a sentence on the contract for each. §"Soundness Enforced" picked up two new entries (10 and 11). §"Remaining gaps" dropped 3, 5, 6 and renumbered.

Three gaps left in mep-0003: pattern shapes at parse time, ImportStmt.Lang registry validation, ExprStmt useless-expression diagnostic, and the FunExpr default return question (which is bigger, may not fit this PR). Working pattern shapes next.

tamnd:
Cycle 3 done. Pattern shape validation now runs at parse time.

The AST has always admitted arbitrary expressions in pattern position because MatchCase.Pattern is just `*Expr`. MEP-5 §"Inference for matches" enumerates four pattern shapes; the runtime fall-through quietly treated everything else as an equality test against the evaluated value, which is more like an `if` than a `match`.

New `validatePatternShape` in parser/normalize.go enforces the four shapes structurally: literal, bare identifier, parenthesised pattern, or a variant call whose args are themselves patterns. Anything else gets P063. The check runs as part of the normalisation pass we already had wired up.

Fixtures cover the three obvious wrong shapes — `1 + 2`, `xs[0]`, `p.x`. mep-0003 picks up a new enforced invariant (item 12) and loses the corresponding gap. P063 is now part of the conformance diagnostic-codes list alongside P062 and P070.

Three gaps left in mep-0003: ImportStmt.Lang registry, ExprStmt useless-expression, and FunExpr default return. ImportStmt.Lang registry next.

tamnd:
Cycle 4 done. ImportStmt language is now validated at parse time.

The grammar lets you write any identifier between `import` and the path string. The runtime later resolves unknown values to "module not found" or a fmt.Errorf deep in the build; a typo like `import pythn "math" as math` had no positioned diagnostic anywhere.

Added a known-languages registry to parser/normalize.go covering everything the codebase actually consumes: go, python, typescript, zig, lua, clj, clojure. Anything else is rejected as P064 with a position. The hint lists the legal set so the user can see "did I mean python?" right away.

mep-0003 picks up invariant 13 and drops the corresponding gap. P064 added to the conformance diagnostic-codes list.

Two gaps left: ExprStmt useless-expression diagnostic (next), and FunExpr default return — the FunExpr one really wants a `unit` type which is its own design decision, may not fit this PR. ExprStmt next.

tamnd:
Cycle 5 done. ExprStmt useless-expression diagnostic landed as P065.

The rule: an expression used as a statement must reach a call/fetch/save/load/generate node along an unconditional path that doesn't cross into a function literal body. The crossing rule matters: `fun() => print(1)` as a statement is still useless because the lambda is a value, not a side effect, even though its body looks busy. Match and if branches descend because evaluation is conditional but not deferred.

Two existing fixtures had to be tweaked:
- types/errors/not_indexable: `x[0]` became `let y = x[0]` so T018 still fires.
- types/valid/doom_move: was using `doom.Player{x:1.0, ...}` which the grammar doesn't actually accept as a struct literal (StructLiteral requires a bare Ident, not a Selector). The "var p = doom.Player" + map literal next to it was being parsed as two unrelated statements. P065 caught it. Replaced with code that actually exercises the doom import.

mep-0003 picks up invariant 14. P065 added to the conformance diagnostic-codes list.

One gap left: FunExpr default return. That one's blocked on introducing a `unit` type so the AST can record "no return" distinctly from "infer". Bigger design decision and probably out of scope for this PR. I'll write it up as a remaining-future-work note in mep-0003 and consider this round done unless you want me to push on the unit type.

tamnd:
Done with the last MEP-3 gap (FunExpr default return). Went with the rename you mentioned: types.VoidType -> types.UnitType, and String() now returns "unit" instead of "void".

Scope:
- types/check.go, types/infer.go - type def, builtin signatures, FuncType.String() now suppresses "unit" return (used to suppress "void"). Default for FunExpr/FunStmt with nil Return resolves to UnitType{}, which closes the gap MEP-3 was tracking.
- All callers in interpreter, vm, and transpilers (go/swift/ex/kt/pas/rs/hs/lua/ocaml/fs) follow the rename.
- Target-language emissions of "void" (Go func returns, Java method returns, C/C++/C#/Dart keywords) stay as-is. Those are the target's keyword, not Mochi's type name.
- archived/ left untouched.

MEP-3 changes:
- New invariant 15 in §Soundness Enforced documenting the canonical unit resolution rule for FunExpr.Return == nil.
- §Remaining gaps now reports no known soundness gaps in this revision.

Tests: short suite green across types, parser, interpreter, runtime, golden, tools. Build clean. CI running.

That closes round 5 of the gap-fix work on this PR. With this in, MEP-3 has no remaining open gaps - all five originally listed have landing positioned diagnostics or contract docs:
- LogicCond / AgentBlock / PostfixOp positions (Cycle 2)
- Match-case pattern shape: P063 (Cycle 3)
- Import language registry: P064 (Cycle 4)
- ExprStmt observability: P065 (Cycle 5)
- FunExpr default return: canonical UnitType (this)

Issue #21222, opened 2026-05-11
MEP-4: sync type system doc to current state and close known gaps


tamnd:
Tracking issue for the MEP-4 (Type System) doc work.

## Scope

MEP-4 is the type system MEP. It currently has stale references (the VoidType -> UnitType rename from #21221 isn't reflected) and a handful of acknowledged soundness gaps. Use this issue to track the work on:

- Sync doc to current code (kind table, line refs).
- Complete audit and rewrite in PEP-style structure.
- Address soundness gaps that have separate tracking issues:
  - Pure flag not enforced (#21188)
  - List + list silently widens to [any] (#21187)
  - Missing map key returns zero value (#21186)

Out of scope for now: introducing tuple kind, option propagation, parametric polymorphism. Those are MEP-10 / MEP-12.

PR: #21223

tamnd:
doc shipped in #21223, soundness sub-issues #21186/#21187/#21188 all closed. closing the umbrella.

Issue #21186, opened 2026-05-08
Wire OptionType through inference for missing map key access


tamnd:
Part of #21190.

## Background
MEP 4 (Pinned Decisions) documents the chosen behaviour for `m[k]` when key `k` is absent:

> `m[k]` should infer to `OptionType{V}`, not the zero value of `V`.

This is consistent with MEP 10 A2 (`null` → `OptionType`) and MEP 10 C1 (`T?` shorthand).

## Current behaviour
Today `m[k]` infers to `V` (the map's value type) and the runtime silently returns the zero value of `V` for absent keys. For non-zero-bearing types this is a progress violation.

The current behaviour is pinned by `tests/types/valid/missing_map_key.mochi`.

## Desired behaviour
- Checker infers `m[k]` as `OptionType{V}`.
- Callers must unwrap the option before using the value as `V`.
- Runtime produces the absent variant of `OptionType` for missing keys.

## Work
1. In `inferPostfixType` (`types/infer.go`), change `case MapType: t = tt.Value` to `t = OptionType{Elem: tt.Value}`.
2. Move `tests/types/valid/missing_map_key.mochi` to `tests/types/errors/` and add a `.err` snapshot showing the required unwrap.
3. Add `tests/types/valid/missing_map_key_unwrapped.mochi` demonstrating the correct unwrap pattern.

Blocked on MEP 10 A2 (option type wired through inference).

tamnd:
Re-opening as part of the MEP-4 §6 defer-tracking audit. This issue was closed in the 2026-05-08 batch close when the original umbrella #21190 closed, but the work (wire OptionType through inference for missing map key access) is still pending and is the canonical destination MEP-4 references for problem P5. Blocked on MEP-10 A2 (null → OptionType wiring) per MEP-10 §A2.

tamnd:
landed via 8079448684 and mep-0016 R8 (map index returns V?), closing

PR #21223, opened 2026-05-11
mep-0004: sync type system doc to current state

tamnd:
## Summary

- Updates the §Kinds present today table for the VoidType -> UnitType rename that landed in #21221, and cross-links the rename rationale to MEP-3 invariant 15.
- Softens hard-coded source line ranges in MEP-4 to "around line N" so the doc does not silently stale-mark when types/check.go or types/infer.go grow.

This is the opening commit for the MEP-4 doc-sync work tracked in #21222. Follow-up commits in this PR will start chipping at the soundness gaps MEP-4 enumerates.

## Test plan

- [ ] CI green (build + tests)
- [ ] Docusaurus build clean

### DIFF
--- a/website/docs/mep/mep-0004.md
+++ b/website/docs/mep/mep-0004.md
@@ -222,7 +222,7 @@ The current design has several known problems. They are listed below without rig
 
 **11. `UnionType.Variants` has no stable iteration order.** The text claims declaration order is preserved, but the field is a Go map. Downstream consumers that need stable order (the formatter, the C# emitter's switch table) reconstruct it from the parser tree. The fix is to add an `Order []string` parallel to `StructType.Order`.
 
-**12. `StructType.Methods` ties `types` to `parser`.** `Method.Decl` is a `*parser.FunStmt`. Any future split of the AST into its own package, as [MEP 3](/docs/mep/mep-0003) hints at, has to thread back through `types`. The fix is to keep only the method signature and a body handle, not the AST node.
+**12. `StructType.Methods` ties `types` to `parser`.** `Method.Decl` is a `*parser.FunStmt`, so the type system module depends on the parser AST. The fix is to keep only the method signature and an opaque body handle, not the AST node. [#21249](https://github.com/mochilang/mochi/issues/21249).
 
 **13. `Variadic` is a bool, not a varargs element type.** Nothing prevents a `FuncType` with `Variadic: true` and an empty `Params`. The convention "the last parameter is the variadic element" is enforced by the surface grammar, not by the type. The fix is to encode varargs as a separate field, or to split `Params` into a fixed prefix and a single variadic element.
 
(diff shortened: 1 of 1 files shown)


tamnd:
Pushed the PEP-structured rewrite plus the full design-problem audit you asked for.

The doc went from ~195 lines to ~464. New shape, top to bottom:

- Header now carries Requires (MEP 2, MEP 3) and Informs (MEP 5, MEP 10, MEP 11, MEP 12) so the cross-MEP graph is explicit.
- Terminology section pins the PLT vocabulary (kind, nominal, structural, top, unification, judgement) so the rest of the spec can use it without re-introducing.
- Specification §3 is split into 10 numbered subsections covering Type interface, kind inventory (six tables: primitive, structural, nominal, function, polymorphism, top), what is NOT a kind, the typing context, equality, unification, the numeric tower (three coexisting relations: equality, unify, operator widening), operator typing as a normative table, type formation, and a formal typing judgement with inference rules.
- Soundness Properties section states Progress and Preservation qualified by the "any" perimeter, plus what is unenforced.
- Known Problems §6 is the audit. 28 problems in six buckets:
  * §6.1 Soundness (8): map zero, list+list widen, Pure advisory, any wildcard, null as any, unknown simple type as any, OptionType not produced, cross-kind comparison.
  * §6.2 Coherence (3): equalTypes vs unify disagree on int/int64, unify mixing equality and subtyping, equalTypes symmetric in spelling but enumerated manually.
  * §6.3 Representation (6): StructType.Order/Fields redundant, UnionType.Variants unordered, StructType.Methods cycle with parser.FunStmt, *TypeVar as the only pointer kind, Variadic as bool not as varargs-elem type, reflect.DeepEqual fallback fragility.
  * §6.4 API (4): Env mixing static and runtime, Copy flattening parent chain, SetFunc lowercasing names, ExprType total on nil.
  * §6.5 Surface ergonomics (5): no alias kind, no tuple kind, no row variable, unit not surface, no first-class type.
  * §6.6 Doc (2): line numbers go stale, no formal subtype written down.
- Proposed Enhancements §7 ranks the next five in-scope follow-ups in priority order so we know where to push next.
- Pinned Decisions preserves the map-key contract and the void to unit rename rationale.
- References split into PLT (Pierce TAPL, Harper PFPL, Cardelli & Wegner taxonomy, Wadler expression problem, Damas-Milner, Reynolds) and cross-MEP.

Tests still green on types and parser. Pure doc commit, no code change.

Next pass on this PR (per the priority list in §7): reject unknown simple type names in resolveTypeRef as a new T0xx, and tighten ==/!= on incompatible kinds. Both are small surgical changes.

tamnd:
Took another pass on the prose. Re-read PEP 484, PEP 8, PEP 612 to recalibrate the voice, then rewrote MEP 4 to match.

Changes from the prior version:
- dropped the §3.x section numbering, headings on their own are enough
- collapsed the 28-problem six-category audit into a flat numbered list of 22 items, with a one-paragraph recommendation at the end of which five are worth doing soon
- cut the tables down to two (the arithmetic operator matrix and the reference-implementation pointer table)
- removed the 'How to Teach This' section, the italics-as-emphasis pattern, the 'Severity:' tags, and the 'in scope / out of scope' bureaucracy
- rewrote the prose to read like a person describing the system, varied sentence rhythm, no triplet enumerations, no 'load-bearing' / 'deliberate' / 'first-class' fillers

Same content, much less AI-flavoured.

tamnd:
Audit pass over §Problems just landed in #21252. Before this merges, three cross-references in the new §Problems list are blank ("follow-up MEP", "MEP-3 hints at", "sweep touches"). Suggested edits so the rendered doc points at live trackers:

- **P5** currently reads "tracked in #21186". Issue was prematurely closed in the 2026-05-08 batch close and has been re-opened, so the existing reference is now correct again. No edit needed.
- **P12** currently reads "as MEP-3 hints at". MEP-3 actually argues against a separate ast package (mep-0003.md around line 297). Suggest: replace the MEP-3 reference with a pointer to #21249 and reword the motivation as "decouple the types package from the parser AST".
- **P16** currently has no tracker reference. Suggest: add a pointer to #21250.
- **P20** currently reads "is a follow-up MEP". Suggest: replace with a pointer to #21251, or leave the MEP-stub language and add the issue link alongside.

All three new issues (#21249, #21250, #21251) cite this PR as the canonical source for the §Problems numbering.

Issue #21364, opened 2026-05-12
docs: defer MEP-12.5 explicit type argument syntax


tamnd:
Defer the last open MEP-12 step. The grammar change introduces </> ambiguity with comparison expressions and requires a dedicated parser design pass. Inference works without explicit type arguments; the syntax is opt-in sugar. Tracked as a follow-up parser MEP. Closes task #90.

tamnd:
PR: #21365

tamnd:
closed by #21365, closing

PR #21365, opened 2026-05-12
docs/mep-0012: defer MEP-12.5 explicit type argument syntax

tamnd:
Mark MEP-12.5 deferred. The grammar change requires careful disambiguation of </> against comparison expressions and is opt-in sugar over working inference. Tracked as a follow-up parser MEP.

Tracking: #21364


tamnd:
tracking: #21364

Issue #21366, opened 2026-05-12
cleanup: split parser AST out of parser.go


tamnd:
Pure code motion. parser/parser.go was 734 lines mixing the lexer, every AST type, and the Parse entry point. Splitting AST defs into parser/ast.go leaves parser.go focused on parse entry + lexer.

Tracks CLEAN-8 from the post-MEP-1-12 cleanup sweep.

tamnd:
PR: #21367

tamnd:
closed by #21367, closing

PR #21367, opened 2026-05-12
parser: split AST type defs into ast.go

tamnd:
Pure code motion. parser.go drops from 734 lines to 108 (lexer + Parse entry only); ast.go takes 630 lines of AST defs with their Capture/Name helpers. No behaviour change; full suite passes.

Tracking: #21366


tamnd:
tracking: #21366

Issue #21368, opened 2026-05-12
cleanup: split types/check.go monolith


tamnd:
types/check.go was 3705 lines and 61 top-level funcs mixing kinds, unify, resolve, statement walkers, and expression walkers. Split into 5 files for reviewability. Tracks CLEAN-5 and CLEAN-13.

tamnd:
PR: #21369

tamnd:
closed by #21369, closing

PR #21369, opened 2026-05-12
types: split check.go monolith into 5 files

tamnd:
Pure code motion. check.go (3705 lines) becomes:
- kinds.go (245): Type interface + kind structs + Subst
- unify.go (257): unify algorithm
- resolve.go (125): resolveTypeRef
- check.go (1121): Check entry + statement walkers
- check_expr.go (2000): expression walkers + alias helpers

No behaviour change; full suite passes.

Tracking: #21368


tamnd:
tracking: #21368

Issue #21370, opened 2026-05-12
cleanup: split runtime/vm/vm.go monolith (pass 1)


tamnd:
vm.go is 9319 lines and 124 funcs. First pass extracts opcode defs, program structures, and error formatting. The eval loop and the compiler remain in vm.go pending further splits. Tracks CLEAN-10.

tamnd:
PR: #21371

tamnd:
closed by #21371, closing

PR #21371, opened 2026-05-12
runtime/vm: extract ops, program, error into separate files

tamnd:
Pure code motion. First slice of vm.go split (CLEAN-10). vm.go drops 570 lines. Eval loop and compiler stay in vm.go pending follow-up passes.

- ops.go (297): Op type + opcode constants + Op.String
- program.go (216): Instr/Function/Program/closure/Disassemble
- error.go (70): VMError + formatting

Full test suite passes.

Tracking: #21370


tamnd:
tracking: #21370

Issue #21410, opened 2026-05-12
MEP-14: pin left/outer join cardinality and correct option-unwrap attribution


tamnd:
Pin the cardinality of `left join` and `outer join` so the eventual `Option<T>` retypy is a visible diff against fixtures, and fix the doc misattribution that pointed at MEP-10 A2 (closed) as the blocker.

## Scope

- Add `tests/types/valid/query_join_left_unmatched_null.{mochi,golden}` (T032/T034 inputs; result row count pinned).
- Add `tests/types/valid/query_join_outer_both_sides.{mochi,golden}` (disjoint sides; row count pinned).
- Update `website/docs/mep/mep-0014.md`:
  - Move both fixtures from "Pinning gaps" into "Existing".
  - Rewrite the left-join wart paragraph: the blocker is option-unwrap discipline (auto-wrap on assignment, auto-unwrap or pattern match on access), not MEP-10 A2.
  - Update Open Questions and the `join` clause description to match.

## Out of scope

- Implementing the `Option<T>` retypy itself. Needs the unwrap rules first.

tamnd:
Tracking PR: #21411

tamnd:
closed by #21411, closing

PR #21411, opened 2026-05-12
MEP-14: pin left/outer join cardinality fixtures

tamnd:
Closes part of #21410.

## Summary

- Add `tests/types/valid/query_join_left_unmatched_null.{mochi,golden}` and `query_join_outer_both_sides.{mochi,golden}` to pin the result row count for left and outer joins.
- Update `website/docs/mep/mep-0014.md`:
  - Move both fixtures from "Pinning gaps" into "Existing".
  - Rewrite the left-join wart so it points at the actual blocker (option-unwrap discipline) instead of MEP-10 A2, which is closed.
  - Tighten the matching Open Question and `join` clause description.

## Test plan

- [x] `go build ./...`
- [x] `go test ./types/ -run TestTypeChecker_Valid -count=1`

### DIFF
--- a/tests/types/valid/query_join_left_unmatched_null.golden
+++ b/tests/types/valid/query_join_left_unmatched_null.golden
@@ -1,0 +1,1 @@
+✅ Type Check Passed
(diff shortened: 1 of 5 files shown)


tamnd:
tracking issue: #21410

Issue #21448, opened 2026-05-13
mep-0015: stage 3a widen T044 to name effect labels


tamnd:
Tracking issue for MEP-15 effect system stage 3a.

## Scope

Widen `T044` (impure call in `where` / `having` predicate) so the diagnostic names the offending effect labels instead of saying just "impure". The help text guides the user to pre-compute or hoist the effectful call.

Before:

> error[T044]: impure call to `print` is not allowed in `where` predicate

After:

> error[T044]: call to `print` produces effect(s) io, not allowed in `where` predicate

## Implementation

- `firstImpureCall` and its helpers now return the callee's `EffectSet` along with name + position
- `errImpurePredicate` takes the effect set and renders it in the message
- Golden in `tests/types/errors/impure_in_where_rejected.err` updated

## Follow-ups (separate PRs)

- Stage 3b: plumb `FunExpr.Effects` through type checker so `T064` / `T065` fire on anonymous functions
- Stage 3c: reserve `T066` for effects in pure positions (const, struct field defaults)

tamnd:
pr: #21449

tamnd:
closed by #21449, closing

PR #21449, opened 2026-05-13
mep-0015: stage 3a widen T044 to name effect labels

tamnd:
Tracks #21448.

## Summary

- Widen `T044` so it names the offending effect labels: `call to \`print\` produces effect(s) io, not allowed in \`where\` predicate`
- Help text now suggests pre-computing or hoisting the effectful call outside the query
- `firstImpureCall` and helpers return the callee's `EffectSet` along with name + position

## Test plan

- [x] `go test ./types/` (including `TestTypeChecker_Errors/impure_in_where_rejected`)
- [x] `go build ./...`


tamnd:
tracking: #21448

Issue #21452, opened 2026-05-13
website: /docs/mep is 404 and sidebar misses new MEPs


tamnd:
Two related issues on the docs site:

1. https://mochi-lang.dev/docs/mep returns 404. There is no index page in `website/docs/mep/`, so the route only resolves for the individual `mep-NNNN.md` files.
2. The MEP sidebar is hand-maintained in `website/sidebars.js`. MEP-15 and MEP-16 had already shipped to disk but were missing (15 was eventually added; 16 was never added). New MEPs will keep silently dropping off.

## Fix

* Generate `website/src/data/meps.json` from the frontmatter of every `mep-NNNN.md` file in the docs tree. Wire it into `npm run prestart` and `npm run prebuild` so the manifest cannot drift.
* Add `website/docs/mep/index.mdx` at `slug: /mep` that imports the manifest and renders a categorized table of every MEP, with status pills and descriptions.
* Add the missing `mep-0016` entry to the hand-curated sidebar groups (and pin the new index page at the top so users land on the catalogue first).

tamnd:
pr: #21453

tamnd:
closed by #21453, closing

PR #21453, opened 2026-05-13
website: auto-generated MEP index page

tamnd:
Tracks #21452.

## Summary

- New `/docs/mep` index auto-generated from every `mep-NNNN.md` frontmatter
- `prestart` / `prebuild` hooks regenerate `src/data/meps.json` so the catalogue cannot drift
- Sidebar gains the previously missing `mep-0016` and pins the new index page at the top

## Test plan

- [x] `npm run build` (prebuild script writes 17 MEP entries, build succeeds)
- [x] `build/docs/mep.html` exists (previously 404) and references all 17 MEPs

### DIFF
--- a/website/docs/index.mdx
+++ b/website/docs/index.mdx
@@ -38,7 +38,7 @@ The [Mochi Manual](/docs/manual/) documents the language in depth.
 
 ## Design proposals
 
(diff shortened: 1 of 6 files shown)


tamnd:
tracking: #21452

PR #21455, opened 2026-05-16
mep-0015: stage 3c reserve T066 for pure positions

### LINKED ISSUE #21454: mep-0015: stage 3c reserve T066 for pure positions
Tracks the Stage 3c sub-task of MEP-15.

MEP-15 reserves diagnostic **T066** for "effect not allowed in this context" — the gate for future pure positions (`const` declarations, struct field defaults, compile-time eval, type-level args). The MEP explicitly says "Stage 1 does not raise T066; the code is reserved."

This task wires up the reservation so the code is callable the moment those surfaces land:

- add `T066` to the `Errors` map in `types/errors.go`
- add `errEffectInPurePosition(pos, context string, effects EffectSet)` helper
- unit test that asserts the rendered message + help so the contract is pinned
- mark T066 as **shipped (reserved, no firing site yet)** in the MEP-15 status table and migration staging
- no fixture, no parser change — `const` and struct field defaults do not exist yet

tamnd:
Closes #21454.

Stage 3c of MEP-15. Registers diagnostic **T066** ("effect not allowed in this context") so the message contract is locked in before the consuming surfaces (`const` declarations, struct field defaults) land.

## What ships
- `types/errors.go`: `T066` template + `errEffectInPurePosition(pos, context, effects)` constructor.
- `types/effects_test.go`: `TestT066_ReservedForPurePositions` renders the template and asserts the message text, so wording is pinned even though no production code emits it yet.
- `website/docs/mep/mep-0015.md`: status table flips T066 from open to closed (reserved); migration staging gains Stage 3c entry; diagnostics section spells out the no-firing-site state.

## What does not ship
- No firing site. `const` and struct field defaults do not exist yet; T066 will be wired into them when those features land.
- No fixture, no parser change.

## Test plan
- [x] `go test ./types/ -run TestT066_ReservedForPurePositions` passes.
- [x] Full `go test ./types/` green.
- [x] `go vet ./types/` and `go build ./...` clean.

### DIFF
--- a/types/effects_test.go
+++ b/types/effects_test.go
@@ -4,6 +4,8 @@ import (
 	"mochi/parser"
 	"mochi/types"
 	"testing"
+
+	"github.com/alecthomas/participle/v2/lexer"
 )
 
 func TestEffectSet_Basics(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tamnd:
Tracks #21454

PR #21457, opened 2026-05-16
mep-0015: stage 3d HOF callback effect propagation

### LINKED ISSUE #21456: mep-0015: stage 3d HOF callback effect propagation
Tracks Stage 3d of MEP-15.

**Problem.** MEP-15 E5 specifies that a HOF call inherits the effects of its function-typed arguments when those arguments are called. Currently `exprEffects` for a closure argument returns `{}` (construction is pure), so `apply(xs, fun(x) { print(x); x })` is incorrectly typed as pure even though it invokes an `!io` callback.

**Fix.** Add `callableEffects(e, env)` to `types/effects_infer.go`:
- Detects the "bare callable" pattern: an expression that IS a FunExpr literal, or IS a simple identifier of function type.
- Returns the effects that would be incurred by *calling* that value (not just constructing it).

Wire it into both `primaryEffects` (direct calls `f(args)`) and `postfixEffects` (postfix calls `obj.method(args)` / `fn(args)`). For each argument, union in both `exprEffects` (eval cost) and `callableEffects` (call cost, if the arg is a function).

**Also in this PR.** Sync the MEP-15 status table: many rows still show **open** for work that shipped across Stages 1–3c.

tamnd:
Closes #21456.

Stage 3d of MEP-15 — the last non-deferred open item.

## Problem

MEP-15 E5 says a HOF call site should inherit the effects of its function-typed arguments when those arguments are called. Before this PR, `apply(fun(x:int):int{ print(x); x }, 5)` was incorrectly typed as pure: `exprEffects` for a FunExpr argument returns `{}` (closure construction is pure), so the `io` effect in the callback body never reached the enclosing call.

## Fix

New `callableEffects(e, env)` in `types/effects_infer.go`:
- Detects "bare callable" arguments: a FunExpr literal, or a plain identifier of function type.
- Returns the effects that *calling* that value would incur (vs. just constructing it).

Wired into both `primaryEffects` (`f(args)` direct calls) and `postfixEffects` (`obj.method(args)` / postfix calls) — for each argument, union in both `exprEffects` (evaluation cost) and `callableEffects` (call cost).

## Approximation note

The propagation is conservative: any function-typed argument is treated as-if-called, even for HOFs that only store their argument. This is a sound over-approximation (no false negatives; occasional false positives for store-only HOFs). The MEP's hidden row variable is implemented without requiring row types in the surface or the stored `FuncType`.

## Also

Syncs the MEP-15 status-at-a-glance table: all Stages 1–3d rows are now marked closed, with the stage number attached.

## Test plan
- [x] `TestHOFCallbackEffects`: three sub-cases (FunExpr literal `!io`, named function `!io`, pure FunExpr literal)
- [x] Full `go test ./...` green

### DIFF
--- a/types/effects_infer.go
+++ b/types/effects_infer.go
@@ -182,7 +182,7 @@ func postfixEffects(p *parser.PostfixExpr, env *Env) EffectSet {
 	for _, op := range p.Ops {
 		if op.Call != nil {
 			for _, a := range op.Call.Args {
-				eff = eff.Union(exprEffects(a, env))
+				eff = eff.Union(exprEffects(a, env)).Union(callableEffects(a, env))
 			}
 			if p.Target != nil && p.Target.Selector != nil && len(p.Target.Selector.Tail) == 0 {
 				if t, err := env.GetVar(p.Target.Selector.Root); err == nil {
@@ -212,7 +212,7 @@ func primaryEffects(p *parser.Primary, env *Env) EffectSet {
 			}
 		}
 		for _, a := range p.Call.Args {
-			eff = eff.Union(exprEffects(a, env))
+			eff = eff.Union(exprEffects(a, env)).Union(callableEffects(a, env))
 		}
 		return eff
 	case p.Group != nil:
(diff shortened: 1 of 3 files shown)


tamnd:
Tracks #21456

Issue #21460, opened 2026-05-16
mep-0005: close P18 (builtins parametric)


tamnd:
MEP-5 §Problem 18 audit shows the collection builtins were already retyped to parametric schemes in the MEP-12.4 batch (#21315, #21333 and friends). The few remaining loose signatures (reverse, len, count, exists, sum) are sound but waiting on row polymorphism / numeric bounds, which is a future MEP-12 extension.

Updating MEP-5 P18 to mark substantively closed.

PR: incoming

tamnd:
PR #21461

tamnd:
done — #21461 merged (mep-5 P18 builtins parametric)

Issue #21258, opened 2026-05-11
MEP-5: rewrite to strict static algebraic type system


tamnd:
Rewrite of MEP-5 (Type Inference) to a strict, static, algebraic specification, modeled on Hindley-Milner.

Key commitments:
- Principal types; never silently widen to any.
- No implicit any: the kind is only produced from an explicit surface annotation.
- No nil/null: nullability is Option[T], failure is Result[T, E].
- Flow-sensitive narrowing in match arms and is-checks.
- Exhaustive coverage of closed unions (gated on MEP-13).
- Let-generalisation, lambda-monomorphism, Algorithm W instantiation (gated on MEP-12).

Doc PR shipping the spec: in flight.

Tracker for the 19 concrete code-level defects called out in MEP-5 §Problems. Each becomes its own issue + PR following the MEP-4 P-number pattern.

Breaking changes: see MEP-5 §Backwards Compatibility for the v0.11 warning / v0.12 error cycle.

tamnd:
PR split landed:

- Doc rewrite: #21259
- P2 (`sum()` element type): #21261 / #21260
- P16 (bool result for comparisons/connectives): #21263 / #21262

Remaining §Problems items (P1, P3-P15 except those above, P17-P19) still pending.

tamnd:
MEP-5 §Problems progress update — code fixes shipped:

- P2 (sum element type): #21261
- P5 (match arm widening): #21267
- P13 (for-loop unknown source): #21265
- P16 (bool result for comparisons/connectives): #21263

Remaining items and their status:

**Tractable next:**
- P3 (empty list/map literals self-type to `[any]`/`{any:any}`) — needs fresh type variables, gated on P11.
- P6 (struct field unknown index returns `any`) — partially addressed by the checker's T026; inferrer fallback remains.
- P14 (`select r` falls through) — iteration var inference must produce a principal type; the right shape needs Algorithm W from MEP-12.
- P19 (`ExprType`/`ExprTypeHint` split) — refactor to a bidirectional pass; design-dependent.

**Bigger / breaking:**
- P1 (silent `any` fallbacks, ~40 paths) — gated on v0.11 warning / v0.12 error deprecation cycle.
- P4 (cast compatibility) — gated on subtype predicate from MEP-11.
- P7 (`nil` at runtime entry points) — runtime work, needs `Option`/`Result` constructor injection at every Go-nil boundary.
[...]

PR #21461, opened 2026-05-16
mep-0005: close P18 (builtins substantively parametric)

tamnd:
## Summary
- Mark MEP-5 §Problem 18 substantively closed
- Doc-only change: notes which builtins were retyped in MEP-12.4 and why reverse intentionally stays loose

## Why
P18 audit against current code shows first/distinct/keys/values/push/append/min/max/collect/concat/reduce are all parametric (`∀α. ...α...`) since #21315 and #21333. The remaining loose signatures (reverse, len, count, exists, sum) are sound but want row polymorphism or numeric bounds to fully tighten.

Refs #21460, #21258


tamnd:
Tracking issue: #21460

Issue #21462, opened 2026-05-16
mep-0005 P3: empty literal fresh tyvar (blocked on MEP-12 Subst)


tamnd:
MEP-5 §Problem 3 audit.

**Currently sound:** annotated empty literals like `let xs: list<int> = []` work via the `elementContext` carve-out in `types/subtype.go` (`assignableAt`), which admits `[any]` flowing into `[int]` only at the literal site.

**Still unsound:** the unannotated case below type-checks today:
```
let xs = []
push(xs, 1)
push(xs, "s")
```
After binding, `xs : [any]` accepts anything forever; no T030 fires.

**Fix requires:**
1. empty literal returns `ListType{Elem: FreshTypeVar("T")}` / `MapType{Key: FreshTypeVar("K"), Value: FreshTypeVar("V")}` instead of `AnyType`
2. `Subtype` / `Assignable` learn a TypeVar rule (currently no case → falls through to false)
3. persistent `Subst` across statements + `generalise` at the let binder

(1) and (3) hang off Algorithm W plumbing tracked in MEP-12. Tagging P3 as **partial** for now and parking the unannotated case behind that work.

Doc update PR: incoming.

tamnd:
PR #21463

PR #21463, opened 2026-05-16
mep-0005: clarify P3 partial status, park unannotated case on MEP-12

tamnd:
## Summary
- Audit MEP-5 §Problem 3 against current code
- Document that annotated empty literals are already sound via the `elementContext` carve-out in `subtype.go`
- Park the unannotated heterogeneous-append case (still unsound) on MEP-12 (a/c need persistent Subst + Generalise)

## Why
P3's original text said "change `inferPrimaryType` for the empty cases and rely on let-generalisation". That fix can't land standalone: switching empty `[]` to a fresh `TypeVar` breaks `let xs: list<int> = []` until `Subtype/Assignable` learn a `TypeVar` rule and the let-binder threads a persistent `Subst`. Both are MEP-12 territory.

## Repro of the still-unsound case
```
let xs = []
push(xs, 1)
push(xs, "s")
```
Currently type-checks; should be T030 once P3 fully lands.

Refs #21462, #21258


tamnd:
Tracking issue: #21462

tamnd:

Issue #21464, opened 2026-05-16
mep-0005 P14: close (query select projection already principal)


tamnd:
Audit shows P14 is already closed. `checkQueryExpr` in `types/check_expr.go` calls `checkExpr` on the select expression (not a fall-through) and wraps the result as `ListType{Elem: selT}`. The originally-loose `any` projections came from generic builtins returning `any`, which was sealed by MEP-10 A1 (Assignable on call args) and MEP-12.4 (parametric TypeVar signatures) per bb4f1d2c87.

Doc update PR: incoming.

tamnd:
PR #21465

tamnd:
done — #21465 merged (mep-5 P14 query select projection principal)

PR #21465, opened 2026-05-16
mep-0005: close P14 (query select projection is principal)

tamnd:
## Summary
- Audit MEP-5 §Problem 14 against current code
- Mark P14 closed: `checkQueryExpr` already produces a principal projection type
- Cross-reference the MEP-10 B6 close (bb4f1d2c87) which sealed the same loose-any source

## Why
P14's original text predates MEP-10 A1 and MEP-12.4. After those landed:
- Query select calls `checkExpr` proper on the projection
- Generic builtin returns are parametric, not `any`
- Only user-declared `: any` returns produce any-typed projections

Probes pass: `from x in [1,2,3] select x + 1` types as `list<int>`; `from x in [1,2,3] select x.unknown` raises T027 cleanly.

Refs #21464, #21258


tamnd:
Tracking issue: #21464

Issue #21466, opened 2026-05-16
mep-0005 P6: close (partial reads consistent)


tamnd:
Audit shows P6 closed. Three partial-read paths now consistent:

1. struct field unknown name → T026 hard error (stronger than Option)
2. map index unknown key → `Option[V]` via MEP-16 N1/N3
3. query `select x["k"]` propagates option to `list<Option<V>>`

Verified: `let ys: list<int> = from x in xs select x["c"]` raises T008 because `[option[int]]` does not flow into `[int]`.

Doc update PR: incoming.

tamnd:
PR #21467

tamnd:
done — #21467 merged (mep-5 P6 partial reads consistent)

PR #21467, opened 2026-05-16
mep-0005: close P6 (partial reads consistent)

tamnd:
## Summary
- Audit MEP-5 §Problem 6 against current code
- Mark P6 closed: all three partial-read paths now consistent

## Why
- struct field unknown name → T026 hard error
- map index unknown key → `Option[V]` (MEP-16 N1/N3)
- query select wrapping a map index → `list<Option<V>>` propagated

Refs #21466, #21258


tamnd:
Tracking issue: #21466

Issue #21468, opened 2026-05-16
mep-0005 P17: bidirectional param types for unannotated lambdas


tamnd:
Closes MEP-5 §Problem 17 for the call-site case.

`checkFunExpr` already plumbed the call-site expected type to the lambda's return; now it also routes missing parameter types through it. Probe:

```
fun apply(f: fun(int): int, x: int): int { return f(x) }
let r: int = apply(fun(x) => x + 1, 5)  // x infers to int from the expected fun(int): int
```

Before: T008 "got string" (the lambda's any-typed params produced a weird inferred signature). After: r == 6.

The fallback to `any` only fires for top-level unannotated lambdas with no enclosing expected type. That residue is the same as P3's unannotated-let-generalisation problem and stays parked on MEP-12 Algorithm W plumbing.

PR: incoming

tamnd:
PR https://github.com/mochilang/mochi/pull/21469

PR #21469, opened 2026-05-16
mep-0005: close P17 (bidirectional param types for unannotated lambdas)

tamnd:
## Summary
- Route missing lambda parameter types through the call-site expected `FuncType` in `checkFunExpr`
- Add `tests/types/valid/closure_bidirectional_param.mochi` to lock the new behavior
- Update MEP-5 P17 to record the closure

## Why
`apply(fun(x) => x + 1, 5)` used to type-error because the lambda's unannotated `x` defaulted to `any`, then the expected `fun(int): int` could not unify. The expected type was already plumbed to `checkFunExpr` (used for the return type); P17 just propagates it into the parameter slots as well. The fallback to `any` still fires for top-level unannotated lambdas with no enclosing expected type (same residue as P3, parked on MEP-12 Algorithm W).

## Test plan
- [x] new fixture passes (`go test ./types/...`)
- [x] no regressions in parser/runtime test suites

Refs #21468, #21258


tamnd:
Tracking issue: #21468

tamnd:

Issue #21470, opened 2026-05-16
MEP-5 P19: collapse ExprType / ExprTypeHint into a shared pass


tamnd:
Tracking issue for MEP-5 Problem 19.

The two public entry points `ExprType` and `ExprTypeHint` in `types/infer.go` duplicated the list/map hint dispatch. Collapse them into a single shared helper so the synthesis pass has one entry point.

Deeper bidirectional propagation (threading the hint through every recursion node) is out of scope here; that work belongs to the Algorithm W layer in MEP-12.

Refs MEP-5 §Problems #19.

tamnd:
PR: #21471

PR #21471, opened 2026-05-16
mep-0005: collapse ExprType / ExprTypeHint into a shared pass (P19)

tamnd:
Closes the last tractable item from MEP-5 §Problems.

## What changed
- `types/infer.go`: `ExprType` and `ExprTypeHint` now both delegate to a single internal `exprTypeWithHint` helper. The duplicated list/map hint dispatch is gone.
- `website/docs/mep/mep-0005.md`: P19 marked closed, with a note that deeper bidirectional propagation through every recursion node remains MEP-12 Algorithm W work.

## Why
P19 in MEP-5 flagged the duplicated hint logic as drift. Two public names are kept because ~12 compiler backends call them, but the implementation is unified.

## Tests
- `go test ./types/...` passes.

Refs #21470


tamnd:

Issue #21474, opened 2026-05-16
mep-0005 P8: load/save/fetch should return result<…, IOError>


tamnd:
Tracking issue for MEP-5 §Problems P8.

Today `load`, `save`, and `fetch` return raw `list<T>`, `unit`, and `T`. MEP-5 §Problems P8 says they need to return `result<…, IOError>` so that the failure surface is visible in the static type and callers must destructure.

## Subtasks
- [x] P8.0 - extract prelude to single file (groups, easy enable/disable)
- [x] P8.1 - add `ResultType` kind, wire through subst/unify/subtype/poly/infercheck/infer
- [x] P8.2 - resolve `result<T,E>` and `option<T>` at the type-ref layer
- [x] P8.3 - register `IOError` closed union (NotFound, PermissionDenied, InvalidFormat, NetworkError, IOFailure)
- [x] P8.4 - load/save/fetch signatures return `result<…, IOError>` with bidirectional Ok unwrap
- [ ] P8.5 - VM runtime Result representation
- [ ] P8.6 - migrate ~149 fixtures using load/fetch + 57 using save
- [ ] P8.7 - migrate transpiler backends, close MEP-5 P8 row in the doc

Paused after P8.4 because P8.6 (fixture migration) is the big rock; resuming later.

PR: TBD

tamnd:
PR: https://github.com/mochilang/mochi/pull/21475 (draft, paused after P8.4)

PR #21475, opened 2026-05-16
mep-0005 P8: load/save/fetch return result<…, IOError> (WIP)

tamnd:
Tracks #21474. Draft, paused after P8.4.

## What's in so far

- **P8.0** prelude refactor. Pulled the inline 250-line builtin registration block out of types/check.go into types/prelude.go, broken into groups (collection / string / numeric / time / meta / io / error-types) so embedders can disable a bundle without surgery.
- **P8.1** `ResultType{Ok, Err}` kind plus subst, occurs, unify, subtype (covariant on both arms), poly free-var walk, infercheck `ContainsAny`, infer `equalTypes`.
- **P8.2** `resolveTypeRefInner` understands `result<T, E>` and `option<T>` as 2- and 1-arg generic surface syntax.
- **P8.3** `IOError` registered as a closed union with five narrow variants: NotFound{path}, PermissionDenied{path}, InvalidFormat{path, reason}, NetworkError{url, reason}, IOFailure{reason}. Variants are also exposed as structs so `match` can name them.
- **P8.4** load -> `result<list<T>, IOError>`, save -> `result<unit, IOError>`, fetch -> `result<T, IOError>` where T flows in from a bidirectional expected type when present (`unwrapResultOk` peels one Result layer off the let's expected type for the fetch's Ok arm).

## What's left

- **P8.5** pick a runtime Result representation in the VM. Two options on the table: new `ValueResult` tag with (ok bool, payload Value), or sentinel `ValueMap{__ok__: v}` / `ValueMap{__err__: v}` mirroring the existing `__group__` convention.
- **P8.6** 149 fixtures use load/fetch and 57 use save. All need to unwrap with match/if let, plus golden re-baseline.
- **P8.7** transpiler backends in compiler/x/* need codegen for the Result shape. Then close the P8 row in docs/specs/mep-0005-error-handling.md.

## Why paused

P8.6 is the rock. Going to come back to it after other work clears.

## Test plan
- [x] `go build ./...` clean after P8.4
- [x] `go test ./types/...` passes
- [x] `go test ./...` passes (fixtures don't tighten on the new Result shape yet, that's P8.6)
- [ ] re-run full suite after P8.5 lands
- [ ] re-baseline goldens after P8.6


tamnd:

Issue #21488, opened 2026-05-16
MEP-20: bring up runtime/vm2 with narrow Value, frame/regs pools, IC, quickening


tamnd:
Bring up `runtime/vm2/` as a clean-slate VM implementing MEP-20 simultaneously with MEP-18 Phase C (Instr.Quick byte) and MEP-19 (polymorphic call IC + tag-specialized quickening).

## Why a new package
MEP-20's Value flip touches ~387 field accesses in runtime/vm plus every Value literal in tools/, bench/, cmd/. An in-place refactor was too risky for a single PR; vm2 lets us validate the layout decisions on real numbers without rewriting every handler.

## What's in vm2
- 24-byte Value (Tag + Num int64 + Ref any), float via Float64bits
- sync.Pool for *frame, per-size []Value slab pools (10 power-of-2 buckets, 4..2048)
- Iterative dispatch loop with explicit *frame stack + callReturn records (no Go recursion through the interpreter — keeps frame pooling effective on deep recursion)
- Instr.Quick byte (MEP-18 Phase C) with tryQuicken/deoptQuicken
- Polymorphic call IC (siteSlots=4, megamorphism after 8 misses)
- Concrete *Closure (no any wrapper)

## Numbers (fib(25))
| build | ns/op | B/op | allocs/op |
|-------|------:|-----:|----------:|
| runtime/vm     | 203,898,861 | 364,739,934 | 515,563 |
| runtime/vm2    |  46,743,623 |       1,576 |      10 |

~4.4x faster, ~232,000x less memory churn, ~50,000x fewer allocs.

## Follow-ups
- port the rest of the MEP-17 bench corpus (IterSum, HofMap, ClosureDispatch, etc.) to vm2 bytecode
- decide whether to migrate the production compiler target from vm to vm2, or fold vm2's pieces back into vm

PR #21489, opened 2026-05-16
mep-0020: bring up runtime/vm2 (narrow Value, pooled frames, IC, quickening)

tamnd:
Refs #21488.

Clean-slate VM at `runtime/vm2/` so we can validate MEP-20's Value layout on real numbers without an in-place refactor of runtime/vm (which would touch ~387 field accesses and the entire bench/tools/cmd surface).

## What's in
- 24-byte Value (Tag + Num int64 + Ref any); float via Float64bits.
- sync.Pool for `*frame`; per-size []Value slabs in 10 power-of-2 buckets (4..2048). Slab pointer cached on the frame so release doesn't escape a fresh `*[]Value` per call.
- Iterative dispatch loop with explicit `*frame` stack + `callReturn` records — no Go recursion through the interpreter, so deep recursion doesn't defeat the pool.
- Polymorphic call IC (siteSlots=4, megamorphism after 8 misses).
- tryQuicken / deoptQuicken on Add/Sub/Mul/Less/Index hot paths from day one.

## Numbers: fib(25)
| build        | ns/op       | B/op        | allocs/op |
|--------------|------------:|------------:|----------:|
| runtime/vm   | 203,898,861 | 364,739,934 | 515,563   |
| runtime/vm2  |  46,743,623 |       1,576 |        10 |

~4.4x faster, ~232,000x less memory churn, ~50,000x fewer allocs.

## Test plan
- [x] `go build ./runtime/vm2/...`
- [x] `go vet ./runtime/vm2/...`
- [x] `go test ./runtime/vm2/bench/ -run TestFib10`
- [x] `go test ./runtime/vm2/bench/ -bench BenchmarkFib25 -benchmem`

## Follow-ups (tracked in #21488)
- port the rest of the MEP-17 bench corpus (IterSum, HofMap, ClosureDispatch, ...) to vm2 bytecode
- decide whether to migrate the production compiler target from vm to vm2 or fold vm2's pieces back into vm

### DIFF
--- a/runtime/vm2/bench/fib_test.go
+++ b/runtime/vm2/bench/fib_test.go
@@ -1,0 +1,78 @@
+package bench
+
+import (
+	"testing"
+
+	"mochi/runtime/vm2"
+)
+
+// fibProgram hand-builds the bytecode for:
+//
+//	fn fib(n int) int {
+//	  if n < 2 { return n }
+//	  return fib(n-1) + fib(n-2)
+//	}
+//	fn main() int { return fib(N) }
+//
+// Layout (no globals):
+//   fib regs: r0=n(param), r1=2, r2=cmp, r3=1, r4=n-1, r5=fib(n-1),
+//             r6=2(const), r7=n-2, r8=fib(n-2), r9=sum
+//   main regs: r0=N, r1=result
+func fibProgram(n int) *vm2.Program {
+	fib := vm2.Function{
+		Name:      "fib",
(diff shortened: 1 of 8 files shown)


tamnd:
Superseded by MEP-21 v2 first-principles redesign (#21495). The 24-byte Value layout here is incompatible with the 8-byte NaN-boxed Cell required by v2 spec. Rebooting from scratch in a new series.

Issue #21599, opened 2026-05-17
mep-0038 §A.6: return-superop family + expanded pair bench suite


tamnd:
Tracking issue for MEP-38 §A.6.

Three pieces:

1. Return-superop family (OpReturnI64K, OpReturnAddSum, OpReturnNewPair). Fuses small return-expression trees into single dispatches. Cuts binary_trees_pair (D=8) from 33 us to 21 us.

2. OpTailCallSelfA3. Fuses the parallel-move schedule for 3-parameter self-tail-calls into one dispatch. Hits every loop(i, n, head) shape in the pair benchmark suite.

3. Two new BG-style pair kernels (pair_swap, pair_walk) demonstrating the arena win on different aspects of pair usage.

Measured (M4, count=3, benchtime=2s):

| Kernel | Ratio (VM2 / Go) |
|--------|----------------:|
| pair_throughput | 0.90x (VM2 faster) |
| pair_swap | 1.74x |
| pair_walk | 1.82x |
| binary_trees_pair (D=8) | 2.55x |
| binary_trees_pair (D=12) | 2.54x |

binary_trees_pair stays at 2.55x because it uses non-tail recursion; full Phase 7 JIT lowering of OpCallA2 + OpReturn is the path to 2x there.

tamnd:
PR: #21600

tamnd:
done — #21600 merged (mep-38 §A.6 return-superop family)

PR #21600, opened 2026-05-17
mep-0038 §A.6: return-superop family + pair bench suite

tamnd:
Refs #21599

## Summary

Three changes that round out the pair-arena benchmark game on MEP-38 §A.6:

1. **Return-superop family**. `OpReturnI64K`, `OpReturnAddSum`, `OpReturnNewPair` collapse the small return-expression trees that dominate the leaf and inductive paths of recursive BG kernels (checkTree, makeTree). binary_trees_pair (D=8, reused VM) drops from ~33 us to ~21 us.
2. **OpTailCallSelfA3**. Fuses the parallel-move schedule and IP rewind of a 3-parameter self-tail-call into one dispatch.
3. **Two new pair-arena BG kernels**: `pair_swap` (tight allocate-and-swap loop) and `pair_walk` (build-then-traverse).

## Measured (M4, count=3, benchtime=2s)

| Kernel | VM2 ns/op | Go ns/op | Ratio |
|---|---:|---:|---:|
| pair_throughput | ~12,400 | ~13,800 | **0.90x** (VM2 faster) |
| pair_swap | ~19,000 | ~10,900 | **1.74x** |
| pair_walk | ~27,000 | ~14,800 | **1.82x** |
| binary_trees_pair (D=8) | ~21,000 | ~8,200 | 2.55x |
| binary_trees_pair (D=12) | ~344,000 | ~135,000 | 2.54x |

Three of the four pair-arena benchmark game kernels meet the 2x merge gate; pair_throughput beats Go outright (the loop body collapses to 4 dispatches per iter).

binary_trees_pair stays at 2.55x because it is the only kernel in the suite that uses non-tail recursion: `return 1 + check(left, d-1) + check(right, d-1)` builds an arithmetic expression across the two recursive calls, so the loop function cannot lower to `OpTailCallSelf` and each node visit pays full `OpCallA2` frame push/pop. The §A.6 fusions removed the body-eval cost (33 us → 21 us); what remains is structural call-frame overhead, the scope of Phase 7 JIT lowering (documented in §A.6 and §B.6 of MEP-38).

## Test plan
- [x] `go test ./runtime/vm2/... ./compiler2/... ./runtime/jit/...` green
- [x] New tests `TestPairSwapKernel`, `TestPairWalkKernel` pass
- [x] `go test -bench=BG_Pair -benchtime=2s -count=3` shows the ratios above
- [x] Existing `binary_trees_pair`, `pair_throughput` kernels still pass and improve

### DIFF
--- a/compiler2/corpus/bg_pair_kernels.go
+++ b/compiler2/corpus/bg_pair_kernels.go
@@ -1,0 +1,125 @@
+package corpus
+
+import "mochi/compiler2/ir"
+
+// BuildPairSwapKernel builds a tail-recursive pair-swap loop. Each
+// iteration reads the two halves of the current pair and constructs a
+// fresh pair with the slots swapped. The chain head escapes through the
+// final return so the swap can't be elided. MEP-38 §A.5: this is the
+// minimum-body pair workload, where every loop step does exactly one
+// arena bump and two projection reads. Go must heap-allocate each
+// swapped pair; vm2 bumps the pair arena.
+//
+// Shape:
+//
+//	main()              = loop(0, N, pair(0, 1))
+//	loop(i, n, head)    = if i >= n then head
+//	                      else loop(i+1, n, pair(head.snd, head.fst))
+//
+// Steady-state body (post-fusion): OpJumpIfLessI64, OpPairFst, OpPairSnd,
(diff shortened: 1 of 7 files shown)


tamnd:
tracking: #21599

Issue #21655, opened 2026-05-18
mep-0039 §6.7: k_nucleotide iteration 2 (bulk super-op + i64 array counts)


tamnd:
Iteration 2 of the BG k_nucleotide kernel under MEP-39 §6.7. Iter 1 left vm2 at 15.99x of Go (~311 ns/iter); the per-iter cost was dominated by ~230 ns of map work (four MapGet/MapSet pairs on a generic boxed map).

This iteration ships mechanism 1 + mechanism 3 from the §6.7 candidate table together:

- `OpKNucleotideRun` bulk super-op that runs the full LCG + cumprob cascade + counts increments inline in Go (same shape as iter 7 super-ops on reverse_complement).
- Counts move from `TMap` to a 20-slot `TI64Array`, so the rolling hash rollup uses register-resident slice indexing rather than a boxed-key map probe.

Targets:
- 2x-vs-Go on N=10000 and N=100000 (macOS arm64, crosslang harness median of 21)
- bit-identical output (723253870 / 1936471392)
- linux re-bench tracked separately

Spec section §6.7 picks up the iter 2 sub-section with theory, mechanism, and bench table.

Tracks PR (link added after open).

tamnd:
Tracks #21656.

tamnd:
done — #21656 merged (mep-39 §6.7 k_nucleotide iter 2)

PR #21656, opened 2026-05-18
mep-0039 §6.7: k_nucleotide iteration 2 (bulk super-op + i64 array counts)

tamnd:
## Summary

- Adds `OpKNucleotideRun` vm2 super-op that drives the entire BG k_nucleotide kernel (LCG + HOMO_SAPIENS cumprob cascade + 1-mer / 2-mer counts) inline in Go, collapsing ~26 dispatches per iter into one.
- Switches counts from a `TMap` to a 20-slot `TI64Array`, eliminating the per-access map hash + bucket walk.
- Adds matching IR op `OpKNucleotideRun`, builder method `KNucleotideRun(counts, n)`, and emit pass-through.
- Rewrites `corpus.BuildKNucleotide` to a 2-function IR (`main` + `summ`); the `loop` and `lookup` helpers and the inline iter-0 bootstrap in `main` are no longer needed.
- Spec §6.7 picks up an iteration 2 sub-section with theory, mechanism, and bench numbers; iteration 1 audit trail retained.

## Result (crosslang macOS arm64, median of 21)

| N      | vm2 (µs) | Go (µs) | vm2 / Go | vs iter 1 |
|-------:|---------:|--------:|---------:|----------:|
|  10000 |      131 |     219 |  **0.60x** | 25.4x faster |
| 100000 |     1152 |    1949 |  **0.59x** | 27.0x faster |

vm2 is now ~1.7x faster than Go on this row, well inside the 2x gate. Linux re-bench on server2 tracked separately.

## Test plan

- [x] `go test ./runtime/vm2/... ./compiler2/...` green
- [x] oracle test `TestCorpusMatchesOracle/bg_k_nucleotide` passes (bit-identical 723253870 / 1936471392)
- [x] `go run ./bench/crosslang -programs bg/k_nucleotide -langs vm2,go -repeat 21` shows vm2 < 2x of Go on both N values
- [ ] Linux re-bench on server2 (tracked separately)

### DIFF
--- a/compiler2/corpus/bg_k_nucleotide_scaled.go
+++ b/compiler2/corpus/bg_k_nucleotide_scaled.go
@@ -6,153 +6,65 @@ import "mochi/compiler2/ir"
 // kernel. The canonical BG k_nucleotide program reads the fasta
 // output (a DNA byte stream) and counts k-length substrings into a
 // hash map, reporting counts for k=1, k=2, and a handful of specific
-// k=3..18 mers ("GGT", "GGTA", ...). Iteration 1 of the cross-lang
-// harness reduces that to the load-bearing inner kernel: an LCG
-// random stream (same parameters as fasta §6.6) drives a HOMO_SAPIENS
-// cumprob lookup that returns a 2-bit code (0/1/2/3 for a/c/g/t);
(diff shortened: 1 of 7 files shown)


tamnd:
tracking issue: #21655

Issue #21661, opened 2026-05-18
mep-0039 §6.11: roll back hard-coded BG kernel super-ops


tamnd:
Track the rollback of the three single-purpose BG kernel super-ops that were merged for k_nucleotide_scaled, mandelbrot, and spectral_norm_scaled.

## Why

The opcodes baked the entire canonical BG kernel into one runtime dispatch arm. The wins came from native Go inside the runtime, not from real VM progress. That is not a fair vm2-vs-Go improvement.

## What changes

- Restores per-step IR for the three corpora.
- Comments out builder, emit, and runtime bodies for OpKNucleotideRun, OpMandelbrotKernel, OpSpectralNormKernel.
- Leaves the opcode enum entries in place with DISABLED markers; runtime traps if encountered.
- Adds MEP-39 §6.11 with rejection rationale and fair-improvement criteria.
- Marks §6.2, §6.4, §6.7 iter-2 entries as REJECTED with pointer to §6.11.

## Follow-up

Future MEP-39 iterations must use generic improvements (interp fusions over IR patterns, typed primitives, regalloc/dispatch tuning, or JIT lowering of typed-op families). The reverse_complement helpers stay for now but OpU8ReverseComplementDNA is flagged in §6.11 as the next candidate to re-examine.

PR: TBD

tamnd:
pr: #21662

tamnd:
done — #21662 merged (mep-39 §6.11 disable hard-coded BG super-ops)

PR #21662, opened 2026-05-18
mep-0039 §6.11: disable hard-coded BG kernel super-ops

tamnd:
## Summary

- Disables the three hard-coded BG kernel super-ops (OpKNucleotideRun, OpMandelbrotKernel, OpSpectralNormKernel). They baked entire canonical BG algorithms into one runtime dispatch, so wins came from native Go in the runtime rather than from real VM progress.
- Restores per-step IR for k_nucleotide_scaled, mandelbrot, and spectral_norm_scaled corpora from the pre-iter-2 commits.
- Adds MEP-39 §6.11 with the rejection rationale, fair-improvement criteria, and the restored iter-1 ratios.

## Behavior

- Builder, emit, and runtime dispatch bodies for the three ops are commented out for audit trail. The opcode enum entries stay with DISABLED markers so the numbering is stable; runtime traps if any of them are encountered.
- The reverse_complement helpers (OpU8FillACGT, OpU8ReverseComplementDNA, OpU8SumI64) are kept since they are byte-transform shapes rather than full algorithms. §6.11 flags OpU8ReverseComplementDNA as the next candidate to re-examine.
- Spec headers in §6.2, §6.4, §6.7 mark iter-2 as REJECTED with pointer to §6.11.

## Tracking

Issue: #21661

## Test plan

- [x] go build ./... clean
- [x] go test ./runtime/vm2/... passes
- [x] oracle bit-identical for the three programs (per-step IR was always correct)
- [x] crosslang bench confirms iter-1 ratios are restored (k_nuc 13.67x/13.85x, mandelbrot 20.36x/22.58x, spectral_norm 51.25x/49.32x vs Go)

### DIFF
--- a/compiler2/corpus/bg_k_nucleotide_scaled.go
+++ b/compiler2/corpus/bg_k_nucleotide_scaled.go
@@ -6,65 +6,153 @@ import "mochi/compiler2/ir"
 // kernel. The canonical BG k_nucleotide program reads the fasta
 // output (a DNA byte stream) and counts k-length substrings into a
 // hash map, reporting counts for k=1, k=2, and a handful of specific
-// k=3..18 mers ("GGT", "GGTA", ...). The cross-lang harness reduces
-// that to the load-bearing inner kernel: an LCG random stream (same
-// parameters as fasta §6.6) drives a HOMO_SAPIENS cumprob lookup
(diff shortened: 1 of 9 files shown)


tamnd:
tracking: #21661

PR #21663, opened 2026-05-18
mep-0039 §6.10: hotfix MDX parse error (bare less-than-equals to 16)

tamnd:
## Summary

Hotfix follow-up to #21662. The previous PR's Build & deploy check failed at line 2457 with an MDX parse error: MDX treats less-than followed by equals as the start of a JSX tag. Tests passed and auto-merge fired before this hotfix landed, so main is currently broken for docs deploy.

The fix matches the project pattern (361ec49d30): rephrase the comparison in prose.

## Test plan

- [x] local docusaurus build succeeds


tamnd:
tracking: #21664

Issue #21665, opened 2026-05-18
mep-0039 §6.12: JIT silently off for FP-heavy BG programs (NumRegs > 7)


tamnd:
Tracking issue for #TBD.

Diagnostic finding: every hot function in mandelbrot, spectral_norm, and n_body exceeds maxJITRegs=7 and is silently rejected by vm2jit, so the 20-50x slowdowns vs Go come from the JIT being entirely off for these programs.

## Probe data

| Program | Function | NumRegs | JIT |
|---|---|---:|---|
| mandelbrot | iterate (inner) | 23 | skip |
| spectral_norm | dot (inner) | 18 | skip |
| n_body | inner | 38 | skip |
| spectral_norm | evalA | 6 | OK |

Of 21 functions across the three programs, only evalA JITs.

## Why

`runtime/jit/vm2jit/compile.go` caps `maxJITRegs` at 7. The arm64 backend maps vm2 register `r` to AArch64 GPR `x(9+r)`, exhausting the AAPCS64 caller-saved temporary window (x9-x15). Going beyond 7 requires callee-saved registers (x19-x28 with STP/LDP prologue) or spill-to-stack.

## Plan

- Phase A (next): reclaim x16/x17 scratch uses, bump to 9 regs
- Phase B: callee-saved x19-x28, bump to 19 regs
- Phase C: spill-to-stack for 20-38-reg functions
- Phase D: compiler-side regalloc improvements to cut NumRegs at source

This PR lands the probe and §6.12 diagnostic only. Phase A starts in the next iteration.

tamnd:
pr: #21666

tamnd:
done — #21666 merged (mep-39 §6.12 JIT coverage diagnostic)

PR #21666, opened 2026-05-18
mep-0039 §6.12: diagnostic probe for JIT coverage gap on FP-heavy BG programs

tamnd:
## Summary

- Lands a build-tagged diagnostic test (`runtime/jit/vm2jit/coverage_probe_test.go`, tag `numregs_probe`) that prints NumRegs and JIT compile status for every function in the FP-heavy BG corpora.
- Adds MEP-39 §6.12 with the finding: every hot function in mandelbrot/spectral_norm/n_body exceeds maxJITRegs=7 and is silently rejected by vm2jit, so the 20-50x slowdowns vs Go come from the JIT being entirely off for these programs.
- Spec captures the AArch64 ABI constraint (x9-x15 = 7 caller-saved temporaries), the three SSA-level reasons NumRegs inflates (tail-recursive parameters, branch-side dead values, intermediate FMA values), and a four-phase plan.

No code-path change to compile or runtime. Phase A starts in the next iteration.

## Tracking

Issue: #21665

## Test plan

- [x] `go test -tags numregs_probe -run TestJITCoverageProbe -v ./runtime/jit/vm2jit/` runs and prints the table in §6.12
- [x] `go build ./...` clean
- [x] local docusaurus build passes (the §6.12 content is MDX-safe)

### DIFF
--- a/runtime/jit/vm2jit/coverage_probe_test.go
+++ b/runtime/jit/vm2jit/coverage_probe_test.go
@@ -1,0 +1,47 @@
+//go:build numregs_probe
+
+// Diagnostic probe: print NumRegs and JIT compile status for every
+// function in the FP-heavy BG corpora. Captured in MEP-39 §6.12.
+//
+// Run with:
+//
+//	go test -tags numregs_probe -run TestJITCoverageProbe -v ./runtime/jit/vm2jit/
+
+package vm2jit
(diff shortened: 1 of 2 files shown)


tamnd:
tracking: #21665

Issue #21671, opened 2026-05-18
MEP-39 §6.15: wire JIT into vm2runner via CompileProgram


tamnd:
The JIT machinery from §6.12/§6.13/§6.14 is unreachable in production. `bench/vm2runner` shells through `emit.Compile` + `vm2.New` and never touches `runtime/jit/vm2jit`. So all the work landed in those sections (cap lifts, callee-saved frames, profitability gate) had zero bench effect: only unit tests via `CompileForceForTest` ever called `Compile`.

§6.15 ships the plumbing: a new `vm2jit.CompileProgram(p)` helper that best-effort compiles every function in `p.Funcs`, silently skipping failures so the interp `OpCall` fast path falls through for non-admitted callees. `bench/vm2runner.main` calls it between `emit.Compile` and `vm2.New` so the timed region sees JIT-resident code.

Coverage on darwin/arm64 jumps from "never" to 8/11 BG programs with at least one JIT'd function. Headline delta is `nsieve` N=1000 -52% and N=10000 -30%. Other programs see single-digit moves or stay within noise; programs with zero coverage (`mandelbrot`, `n_body`, `reverse_complement`) move only by noise because no function admits.

This is plumbing, not policy. The cap stays at 17, the 10:1 profitability ratio stays the same. The next iteration (was "iter 14" in §6.14's "Next" plan, now iter 15) is the JIT-side `OpCall` fast path: when `callee.JITCode != nil` the JIT stays in JIT land instead of round-tripping through Go, which relaxes the 10:1 ratio and admits the cap > 9 candidates.

Tracking PR: mochilang/mochi#TBD

tamnd:
PR up at #21672 — plumbing iteration, nsieve N=1000 cuts from 9757us to 4710us. coverage table + §6.15 in mep-0039.md.

tamnd:
done — #21672 merged (mep-39 §6.15 wire JIT into vm2runner)

PR #21672, opened 2026-05-18
mep-0039 §6.15: wire JIT into vm2runner via CompileProgram

tamnd:
## Summary

- The JIT machinery from §6.12/§6.13/§6.14 was unreachable in production: `bench/vm2runner` shells through `emit.Compile` + `vm2.New` without touching `runtime/jit/vm2jit`. Every vm2 number in those sections was pure interp.
- Add `vm2jit.CompileProgram(p)` that best-effort compiles every `fn` in `p.Funcs` and silently skips failures, leaving `fn.JITCode` nil so the interp `OpCall` fast path falls through.
- Call `CompileProgram` from `bench/vm2runner.main` between `emit.Compile` and `vm2.New` so the timed region sees JIT-resident code.

Coverage post-wire on darwin/arm64: 8/11 BG programs have at least one JIT'd function (vs. 0/11 before). Headline delta is `bg/nsieve` N=1000 -52% (9757us → 4710us) and N=10000 -30% (70477us → 49455us). Other JIT'd programs move by single-digit % at large N (where the JIT'd portion matters) or sit within noise at small N. Programs with zero JIT coverage (`mandelbrot`, `n_body`, `reverse_complement`) move only by noise, as expected.

Tracks #21671.

## Test plan
- [x] `go build ./...`
- [x] `go test ./runtime/jit/vm2jit/... ./runtime/vm2/...`
- [x] `go run ./bench/vm2runner -program=bg_spectral_norm -n=200` returns correct hash
- [x] `go run ./bench/vm2runner -program=bg_mandelbrot -n=200` returns correct hash
- [x] Focused crosslang bench (darwin/arm64, repeat=9, vm2 vs go) on 9 BG programs before/after, recorded as §6.15 delta table

### DIFF
--- a/bench/vm2runner/main.go
+++ b/bench/vm2runner/main.go
@@ -18,6 +18,7 @@ import (
 	"mochi/compiler2/corpus"
 	"mochi/compiler2/emit"
 	"mochi/compiler2/opt"
+	"mochi/runtime/jit/vm2jit"
 	vm2 "mochi/runtime/vm2"
 )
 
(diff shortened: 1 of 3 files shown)


tamnd:
tracks #21671. wire-in only, no policy moves. coverage probe + delta table in §6.15.

Issue #21758, opened 2026-05-19
MEP-40 Phase 6.2d.2.d step 4: JIT-lower OpMapSetI64I64 / OpMapGetI64I64 on ARM64


tamnd:
Tracking issue for MEP-40 Phase 6.2d.2.d step 4: full inline ARM64 lowering of `OpMapSetI64I64` (48 words) and `OpMapGetI64I64` (36 words) so the `maps_fill_sum` corpus kernel admits into the Cell-bank JIT.

## Scope

- Add `mapsBase` to `jitArenaCtx`; new helpers in `runtime/vm3/jit_layout.go` for `vmMap.{nLive,table}` and `mapEntry.{hash,key,value}` offsets + strides.
- Admit `OpMapSetI64I64` / `OpMapGetI64I64` in `checkCellBankAdmissible`; gate `fn.NumRegsI64 <= 4` so caller-saved scratch regs `x4,x13..x17` stay free.
- Emit fixed-size sequences:
  - `Set` (48 words): load-factor preamble with new `cmpShiftLSL` encoder, splitmix64 hash mix, AND mask, linear-probe loop, fill block with `nLive++`. Deopts via `StatusMapGrow` (=3) on cap-overflow.
  - `Get` (36 words): preamble (CBZ on empty table), splitmix64, probe loop, miss block (`xA <- 0`).
- Tests: `TestMapsFillSumKernelsCompile`, `TestMapsFillSumEndToEnd` over `n in {0,1,2,8,32,64,128}`.

## Bench (darwin/arm64 Apple M4)

| kernel | pre step 4 | post step 4 | Go fair | ratio |
| --- | ---: | ---: | ---: | ---: |
| `maps_fill_sum_n128`   | 7906 | 2215 | 5231 | 0.42x (vm3 ~2.4x faster than Go) |
| `lists_fill_sum_n128`  | 165  | 340  | 254  | unchanged (no regression) |

All 11 corpus kernels now sit inside 2x of Go on darwin/arm64.

## Spec

`website/docs/mep/mep-0040.md` step 4 rewritten from `(planned)` to `(landed)` with full instruction breakdown + measured numbers. 6.2d.2.d gate line updated to "Met 2026-05-19".

## Follow-ups

- 6.2d.2.e: replicate on linux/amd64.
- Regrow-and-retry on `StatusMapGrow` (analog of 6.2d.2.b step 2.F), only if a BG kernel surfaces a grow-bound map workload.

[...]

PR #21759, opened 2026-05-19
mep-0040 phase 6.2d.2.d step 4: inline ARM64 lowering of OpMapSetI64I64 / OpMapGetI64I64

tamnd:
Tracks #21758.

## Summary

- Inline ARM64 lowering of `OpMapSetI64I64` (48 words) and `OpMapGetI64I64` (36 words), gated on `NumRegsI64 <= 4` so caller-saved scratch regs `x4,x13..x17` stay free.
- Prologue snapshots `&Arenas.Maps[0]` into `jitArenaCtx.mapsBase` and hoists the per-call map slab byte address into `x20`. Layout offsets exposed via new `runtime/vm3/jit_layout.go` helpers and baked as immediates.
- New `cmpShiftLSL(xn, xm, amount)` encoder fuses the LSL into the load-factor compare; `StatusMapGrow` (=3) added for deopt routing.
- Spec `mep-0040.md` step 4 rewritten from `(planned)` to `(landed)` with the full instruction breakdown + measured numbers; 6.2d.2.d gate line updated to "Met 2026-05-19".

## Bench (darwin/arm64 Apple M4, `-benchtime=2s -count=3`)

`BenchmarkCorpusJITRunner/maps_fill_sum_n128`: `2094..2703 ns/op`, median `~2215 ns/op`.
`BenchmarkGoKernelsFair/maps_fill_sum_n128`: `5089..5863 ns/op`, median `~5231 ns/op`.
Ratio: **~0.42x of Go fair** (vm3 roughly 2.4x faster than Go).
`BenchmarkCorpusJITRunner/lists_fill_sum_n128` unchanged at `329..352 ns/op` (no regression).

All 11 corpus kernels now sit inside 2x of Go on darwin/arm64.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./runtime/jit/vm3jit/... ./runtime/vm3/... ./compiler3/corpus/...`
- [x] `go test ./runtime/jit/vm3jit/... ./runtime/vm3/... ./compiler3/corpus/...` (all green on darwin/arm64)
- [x] `TestMapsFillSumKernelsCompile` gates fill (idx=1) + sum (idx=2) compile
- [x] `TestMapsFillSumEndToEnd` validates correctness over `n in {0,1,2,8,32,64,128}`
- [x] `BenchmarkCorpusJITRunner/maps_fill_sum_n128` measured at ~0.42x of Go fair
- [x] No regression on `lists_fill_sum_n128`


tamnd:
tracking: #21758

PR #21803, opened 2026-05-19
mep-0040 phase 6.3.4.l.3: port reverse_complement + admit OpLookupI64KW (closure-pending)

### LINKED ISSUE #21802: MEP-40 phase 6.3.4.l.3: port reverse_complement to compiler3 + admit OpLookupI64KW (closure-pending)
## Scope

Phase 6.3.4.l.3 of MEP-40: port the reverse_complement BG kernel to compiler3.corpus and wire its JIT bench entry. Admit OpLookupI64KW in the cell-bank ARM64 JIT whitelist so the port lands JIT-compiled rather than interp-only.

## Shape

Single vm3 function, 28 ops, NumRegsI64=6, NumRegsCell=2:

1. OpNewList in/out (lifted by preAllocListPrefix=2 into jitCall).
2. Fill loop: in[i] = bases[i%4], out[i] = 0  for i in 0..n.
3. Revcomp loop: out[n-1-i] = complement[in[i]]  for i in 0..n.
4. Sum loop: sum += out[i]  for i in 0..n.

Tables: I64Tables[0] = {65,67,71,84} (ACGT bytes), I64Tables[1] = 256-entry complement (identity except A<->T, C<->G).

## Generic VM enabler

checkCellBankAdmissible's opcode whitelist gains OpLookupI64KW. Previously only i64-only fns could admit this op; cell-bank kernels with a compile-time lookup table were forced to use OpCmp cascades or fall to interp. ARM64 lowering already handles both hoisted (table-base hoist budget) and cold (movImm64 + LDR Xd, [x16, Xidx, LSL #3]) forms.

## Measured (macOS arm64, post-admit)

| Bench | JIT ns/op | Go ns/op | Ratio |
|-|-|-|-|
| reverse_complement_n1000 | 29,229 | 2,236 | 13.07x |
| reverse_complement_n10000 | 280,782 | 17,144 | 16.38x |

Pre-admit-fix JIT n=1000 was ~52,684 ns (essentially same as interp 50,628 ns); admit fix gets the kernel onto the JIT but does not close the BG gate.

## Verdict: closure-pending

Per-element cost is the bottleneck. Cell-bank list push/get/set route through 16-byte Cell wrappers (~7 ns per op); native Go uses []int64 directly (~0.5 ns per op). The ~14x per-element gap caps this kernel regardless of further opcode-level tuning.

## Next step: Phase 6.3.4.l.4

Add an OpI64Array typed surface (parallel to the OpF64Array surface from Phase 6.3.4.j.5.a/b/c). Migrate reverse_complement (and any future int-list kernel) to the typed surface to close under 2x of Go.

## Composite gate

After this lands: 7/11 BG programs closed under 2x on macOS arm64, 8/11 ported (reverse_complement closure-pending). Linux server2 measurement still outstanding (#85).

tamnd:
Resolves #21802 (port + admit; closure deferred to l.4).

## Summary

- Adds compiler3/corpus.ReverseComplement: 28-op 3-loop fill/revcomp/sum shape over 2 cell-bank lists driven by 2 I64Tables (4-entry bases, 256-entry complement) via OpLookupI64KW.
- Admits OpLookupI64KW in checkCellBankAdmissible's whitelist so cell-bank kernels with compile-time lookup tables JIT-compile (previously i64-only fns admitted this op).
- Wires reverse_complement_n1000 / _n10000 into BenchmarkCorpusJITRunner.

## Measured (macOS arm64)

| Bench | JIT ns/op | Go ns/op | Ratio |
|-|-|-|-|
| reverse_complement_n1000 | 29,229 | 2,236 | 13.07x |
| reverse_complement_n10000 | 280,782 | 17,144 | 16.38x |

Closure-pending: cell-bank list per-element cost (~7 ns through Cell wrappers vs ~0.5 ns for native []int64) caps this kernel ~14x of Go. Phase 6.3.4.l.4 adds an OpI64Array typed surface (parallel to the F64Array surface from j.5.a/b/c) to close under 2x.

## Spec

MEP-40 updated with phase 6.3.4.l.3 section dated 2026-05-20 01:22 GMT+7. Measured table + closure-pending verdict + l.4 scope all documented.

## Test plan

- [x] TestReverseComplementMatchesOracle passes for n in {0,1,2,4,7,16,100,1000,8000}
- [x] go test ./compiler3/corpus/ ./runtime/vm3/ ./runtime/jit/vm3jit/ all green
- [x] BenchmarkCorpusJITRunner picks up both new entries; JIT path executes (JITCode != nil)
- [x] No new em-dashes in spec (count stable at 23)

### DIFF
--- a/compiler3/corpus/corpus.go
+++ b/compiler3/corpus/corpus.go
@@ -32,5 +32,6 @@ func All() []*Program {
 		N_body,
 		SpectralNorm,
 		FannkuchRedux,
+		ReverseComplement,
 	}
 }
--- a/compiler3/corpus/reverse_complement.go
+++ b/compiler3/corpus/reverse_complement.go
@@ -1,0 +1,153 @@
+package corpus
+
+import (
+	"mochi/runtime/vm3"
+)
+
+// ExpectReverseComplement mirrors the cross-lang reverse_complement
+// template: fill an n-entry buffer with the repeating ACGT pattern,
+// reverse-complement into a second buffer (A<->T, C<->G), then return
+// the sum of the output byte values as int64. For n a multiple of 4
+// the result is (n/4) * 287 because the complemented byte set per
+// 4-cycle is {T, G, C, A} = {84, 71, 67, 65}, summing to 287.
+func ExpectReverseComplement(n int64) int64 {
+	if n <= 0 {
+		return 0
+	}
+	in := make([]int64, n)
+	out := make([]int64, n)
+	bases := [4]int64{'A', 'C', 'G', 'T'}
+	for i := int64(0); i < n; i++ {
+		in[i] = bases[i%4]
+	}
+	for i := int64(0); i < n; i++ {
+		c := in[i]
+		var cc int64
+		switch c {
+		case 'A':
+			cc = 'T'
+		case 'C':
+			cc = 'G'
+		case 'G':
+			cc = 'C'
+		case 'T':
+			cc = 'A'
+		default:
+			cc = c
+		}
+		out[n-1-i] = cc
+	}
+	var sum int64
+	for i := int64(0); i < n; i++ {
+		sum += out[i]
+	}
+	return sum
+}
+
+// ReverseComplement: BG `reverse_complement` cross-lang shape port. The
+// template fills `in` with `bases[i%4]` for `bases = "ACGT"`, then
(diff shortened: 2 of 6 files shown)


tamnd:
tracking: #21802

Issue #21854, opened 2026-05-20
Phase 6.3.4.n.4: AMD64 cell+f64 i64 cap 6→7 via R13 remap to admit n_body


tamnd:
## Goal
Lift the cell+f64 admission cap on AMD64 from NumRegsI64 <= 6 to <= 7 so the n_body kernel (NumRegsI64=7, NumRegsF64=8, NumRegsCell=7) gets JIT-compiled on linux/amd64.

## Background
Phase 6.3.4.n.3 admitted cell+f64 fns on AMD64 by repurposing R12 as the regsF64 base (R14 already carries the *jitArenaCtx pointer for cell-bank). The trade-off was one fewer i64 slot, so the cap landed at 6.

n_body needs 7 i64 regs, blocking JIT admission on linux/amd64 even after n.3.

## Approach
Remap i64 slot 6 to R13 when fn is cell+f64. Slot 7's normal home (R13) becomes unavailable in that shape, but slot 6 is now back in play, so the effective cap rises to 7.

- `r2xAMD64` takes `*vm3.Function` and branches on `usesR12ForF64AMD64`
- New helpers `usesR12ForI64AMD64` / `usesR13ForI64AMD64` keep push/pop bookkeeping consistent
- `archCaps` in compile.go raises i64Cap from 6 to 7 for cell+f64 shape

## Verification
- `TestNBodyJITAdmitsAMD64`: asserts n_body JIT-compiles on AMD64 and matches the Go oracle for steps in {0,1,2,5,10,100}
- Full vm3jit + compiler3 test suites pass on darwin/arm64
- Cross-compile to linux/amd64 passes

Bench on server2 (linux/amd64) tracked in PR.

PR: linked below

PR #21853, opened 2026-05-20
vm3jit: AMD64 F64Array + I64Array lowering for cell-bank+f64 (Phase 6.3.4.n.3)

tamnd:
## Summary

MEP-40 Phase 6.3.4.n.3 admits cell-bank+f64 fns on AMD64 by pinning the regsF64 base in R12, then lowers F64Array/I64Array ops so spectral_norm (and later n_body, reverse_complement) escape the interp floor on linux/amd64.

- Move regsF64 base to R12 so R14 remains the *jitArenaCtx pointer for cell-bank fns. Cap NumRegsI64 <= 6 when both Cell and F64 banks are present.
- Add AMD64 cold-form lowering for OpF64ArrayGet/Set/Len, OpI64ArrayGet/Set/Len, and the pre-alloc K-prefix forms of OpNewF64Array / OpNewI64Array.
- Widen checkCellBankAdmissibleAMD64 to admit F64 arithmetic / Const / Mov / Conv / Return ops alongside the array accessors.
- Two follow-up correctness fixes uncovered while landing spectral_norm:
  - movsdLoadXMMFromGPR: emit SIB byte 0x24 when the base GPR is R12 (ModR/M.rm = 100 means "SIB follows" on x86_64 regardless of REX.B), and account for the extra byte in prologueLenAMD64.
  - OpAddF64/SubF64/MulF64/DivF64: when destination A aliases source C (A != B), the naive movsd-then-op shape clobbers C. Route through emitSSEArithAMD64 which uses xmm15 as scratch for non-commutative ops and swaps operand order for commutative ones.

Both fixes are generic backend widenings (not super-ops) and benefit any cell-bank+f64 kernel.

## Validation

- TestF64ArrayJITGetSetAMD64, TestF64ArrayJITLenAMD64, TestSpectralNormJITAMD64
- Full vm3jit + vm3 + compiler3 suites green on darwin/arm64
- BG bench on server2 linux/amd64: spectral_norm n=100 lands at 1.99x of Go (under 2x, closed); n=1000 sits at 2.64x pending the Phase 6.3.4.j.4c LICM follow-up

## Spec

MEP-40 updated with Phase 6.3.4.n.3 section (timestamp 2026-05-20 14:09 GMT+7) covering R12 repurposing, the SIB byte quirk, the SSE A==C aliasing bug, and the spectral_norm bench numbers.

Tracks #21852.

## Test plan
- [ ] CI green on darwin/arm64 and linux/amd64
[...]

PR #21855, opened 2026-05-20
vm3jit: AMD64 cell+f64 i64 cap 6→7 via R13 remap (Phase 6.3.4.n.4)

tamnd:
Tracking issue: #21854
Builds on PR #21853 (Phase 6.3.4.n.3) which is queued for auto-merge.

## Summary
- AMD64 cell+f64 i64 cap raised from 6 → 7 by remapping slot 6 from R12 (now the f64 base) to R13
- Admits the n_body kernel (NumRegsI64=7, NumRegsF64=8, NumRegsCell=7) on linux/amd64
- New `nbody_amd64_test.go` asserts the kernel JIT-compiles and matches the Go oracle bit-for-bit

## Why
Phase 6.3.4.n.3 admitted cell+f64 fns by repurposing R12 as the regsF64 base (R14 already holds *jitArenaCtx in cell-bank), capping i64 at 6. n_body needs 7, blocking JIT on linux/amd64. The free callee-saved R13 slot is the natural home for the remapped slot 6.

## Implementation
- `r2xAMD64(fn *vm3.Function, r uint16)` returns R13 for slot 6 when `usesR12ForF64AMD64(fn)`, R12 otherwise
- New `usesR12ForI64AMD64` / `usesR13ForI64AMD64` helpers keep `numCalleeSavedPushesAMD64`, `prologueLenAMD64`, `emitPrologueAMD64`, `emitEpilogueAMD64`, `epilogueBytesAMD64` in sync with the layout
- `archCaps` in compile.go raises i64Cap from 6 to 7 for cell+f64 fns (slot 7's R13 home is consumed by slot 6, so the effective cap is 7, not 8)
- Admission gate in `lowerAMD64` rejects cell+f64 with NumRegsI64 > 7 with a fresh error message

## Test plan
- [x] `go test ./runtime/jit/vm3jit/...` on darwin/arm64
- [x] `go test ./compiler3/...` on darwin/arm64
- [x] `GOOS=linux GOARCH=amd64 go build ./...` cross-compile
- [ ] Bench n_body on server2 (linux/amd64); attach result + Go ratio in a follow-up comment

### DIFF
--- a/runtime/jit/vm3jit/arena_ctx.go
+++ b/runtime/jit/vm3jit/arena_ctx.go
@@ -97,3 +97,19 @@ func jitArenaCtxPairsBaseOff() uintptr {
 func jitArenaCtxListsBaseOff() uintptr {
 	return unsafe.Offsetof(jitArenaCtx{}.listsBase)
 }
+
+// jitArenaCtxF64ArrsBaseOff is the byte offset of f64ArrsBase within
+// jitArenaCtx. AMD64 OpF64ArrayGetF64/OpF64ArraySetF64 (Phase 6.3.4.n.3)
+// load it from R14+disp32 to recover the F64 array slab base on
+// cell-bank fns.
+func jitArenaCtxF64ArrsBaseOff() uintptr {
+	return unsafe.Offsetof(jitArenaCtx{}.f64ArrsBase)
+}
+
+// jitArenaCtxI64ArrsBaseOff is the byte offset of i64ArrsBase within
+// jitArenaCtx. AMD64 OpI64ArrayGetI64/OpI64ArraySetI64 (Phase 6.3.4.n.3)
+// load it from R14+disp32 to recover the I64 array slab base on
+// cell-bank fns.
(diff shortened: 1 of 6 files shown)


tamnd:
Server2 (linux/amd64) bench, -benchtime=2s:

```
n_body_n100      vm3jit: 95,150 ns/op   |  Go: 29,539 ns/op    |  3.22x
n_body_n10000    vm3jit: 8,623,292      |  Go: 4,355,797       |  1.98x  <- under 2x
```

n=10000 closes; n=100 stays at the per-call JIT entry overhead floor (5 callee-saved push/pop + arena snapshot + status word) which only amortizes meaningfully at the large size. Both numbers are ~19x faster than vm3 interp.

Spec updated in 7fd06b7753.

tamnd:
Superseded. Empty code diff vs main: the Phase 6.3.4.n.4 AMD64 cell+f64 i64 cap 6→7 R13 remap landed on main as commit 23fe05c77e via the n.8 chain. Branch will be deleted by the repo's merge/close hook.

PR #21859, opened 2026-05-20
vm3jit: AMD64 RDX-pin cells.ptr for fannkuch_redux (Phase 6.3.4.n.2.g)

### LINKED ISSUE #21858: Phase 6.3.4.n.2.g: AMD64 RDX-pin cells.ptr for fannkuch_redux n=10000
Track Phase 6.3.4.n.2.g of MEP-40: pin `cells.ptr` in `%rdx` across straight-line spans on the AMD64 cell-bank lowering so List* hot forms after the first refetch skip the `disp8(%rbx)` reload that Phase 6.3.4.n.2.f still pays per op.

## Approach
- Forward dataflow `computeRDXValidAtEntryAMD64`: seed `true` after prologue, invalidate on Div/Mod/ListPush/NewPair/Call*, reset to `false` at every branch target.
- Two emit shapes for `OpListGetI64`, `OpListGetI64K`, `OpListSetI64`:
  - `rdxValid==true`: short form, base=`%rdx`, no reload.
  - `rdxValid==false`: long form, `mov disp8(%rbx), %rdx` first, then SIB load/store.
- Prologue adds 3 bytes (`mov %rax, %rdx`) right after the cells.ptr hoist load.

## Bench (server2, AMD EPYC, -benchtime=3s -count=10, paired same session)
| Phase | vm3jit ns/op | Go ns/op | JIT/Go |
| :---  | ---:         | ---:     | ---:   |
| n.2.f | 1816         | 747      | 2.43x  |
| n.2.g | 1755         | 747      | 2.35x  |

Improvement real (~5% paired, ~20% vs n.2.f's single-rep claim at 3.07x), but not closure. The 2x gate stays open; n.2.h is queued.

## Outcome
Partial: fannkuch_redux_n10000 moves from 3.07x (n.2.f spec claim) to 2.35x steady-state median on linux/amd64. macOS arm64 unchanged (no ARM64 code touched).

PR: TBD

tamnd:
Closes #21858

## Summary
- Pin `cells.ptr` in `%rdx` across straight-line spans on the AMD64 cell-bank JIT lowering so List* hot forms after the first refetch skip the per-op `disp8(%rbx)` reload that Phase 6.3.4.n.2.f still paid.
- Add `clobbersRDXAMD64`, `isListHotFormAMD64`, `computeBranchTargetsAMD64`, and `computeRDXValidAtEntryAMD64` (single linear forward pass with branch-target reset; seeded `true` after the hoisted prologue, invalidated on Div/Mod/ListPush/NewPair/Call*).
- Two emit shapes for `OpListGetI64` / `OpListGetI64K` / `OpListSetI64`: short (base=`%rdx`, no reload) when `rdxValid==true`, long (`mov disp8(%rbx), %rdx` then SIB) when `rdxValid==false`. `byteCountAMD64` mirrors the same branching so `pcMap` stays consistent.
- Prologue adds 3 bytes (`mov %rax, %rdx`) right after the cells.ptr hoist load.
- MEP-40 spec entry for Phase 6.3.4.n.2.g with honest measurement.

## Bench (server2, AMD EPYC, -benchtime=3s -count=10, paired same session)

| Phase | vm3jit ns/op | Go ns/op | JIT/Go |
| :---  | ---:         | ---:     | ---:   |
| n.2.f baseline | 1816 | 747 | 2.43x |
| n.2.g          | 1755 | 747 | 2.35x |

Improvement ~5% paired (or ~20% vs the n.2.f single-rep spec claim of 3.07x). Real, but not closure. The 2x gate stays open for fannkuch_redux_n10000; n.2.h is queued (candidates: magic-multiply IDIV-by-7 constant-fold, widen the pin to the OpListSetI64 short form unconditionally, hoist `cells.len` for the Set-side bounds check).

## Test plan
- [x] `GOARCH=amd64 GOOS=linux go build ./runtime/jit/vm3jit/ ./bench/...` clean on darwin host
- [x] `go test -tags=jit -count=1 ./runtime/jit/vm3jit/...` green on darwin/arm64 (no ARM64 code touched)
- [x] `go test -tags=jit -count=1 ./runtime/jit/vm3jit/...` green on linux/amd64 (server2)
- [x] `BenchmarkCorpusJITRunner/fannkuch_redux_n10000` head-to-head against `BenchmarkGoKernels/fannkuch_redux_n10000` paired in same session, n.2.f reverted vs n.2.g landed


tamnd:
Superseded. Empty code diff vs main: the Phase 6.3.4.n.2.g AMD64 RDX-pin work for fannkuch_redux is subsumed by the n.2.h magic-multiply (commit 7737f90468) plus n.8 closure. fannkuch_redux currently sits at vm3 132us vs Go 146us (N=10000) in the v0.11.0 BG sweep, so the n.2.g pin is no longer the bottleneck it targeted. Branch will be deleted by the repo's merge/close hook.

Issue #22049, opened 2026-05-22
MEP-42 Phase 4.2.28: `in` operator on TypeStr (C target)


tamnd:
Tracking issue for MEP-42 Phase 4.2.28.

## Goal

Light up `needle in haystack` substring containment on TypeStr through `mochi build --target=c`. Before this phase, `applyBinOp` on TypeStr only knew `+`, `==`, `!=`; `if "w" in s` died at lower time with `frontend: operator "in" on str unsupported in MVP`.

## Surface

- New op `OpStrIn(TypeStr, TypeStr) -> TypeBool`, Dispatch (read-only).
- C target: `(strstr(haystack, needle) != NULL)` via `<string.h>`.
- Go target: `strings.Contains(haystack, needle)`.
- Two new fixture pins (positive + negative arm, multi-byte substring).

## Why this closes a user-facing goal

v0.5/string.mochi's `if "w" in s` and `if "z" in s` lines now build on the C target. The full fixture still needs Phase 4.2.29 (`for ch in s` rune iteration) before it pins end to end.

## Limitations

- TypeMapStrI64 `k in m` still rejects (separate phase: `mochi_map_str_i64_has`).
- TypeList / TypeF64Arr / TypeStrArr containment still rejects.
- Empty-needle case returns true on both targets (matches VM and Go semantics).

MEP-42 §10.55 closeout has the full diff shape and limitation notes.

tamnd:
PR: #22050

tamnd:
Landed. Closing.

PR #22050, opened 2026-05-22
mep-0042: Phase 4.2.28 `in` operator on TypeStr (C target)

tamnd:
## Summary

- New `OpStrIn` op (Dispatch, read-only) lights up `needle in haystack` substring containment on TypeStr through the C target.
- C emits `(strstr(haystack, needle) != NULL)`; Go emits `strings.Contains(haystack, needle)`. Both probes are byte-wise so the two targets agree with the VM.
- Two new fixture pins (positive + negative arm, multi-byte substring).

Tracking issue: #22049. MEP-42 §10.55 closeout in the same PR.

## Test plan

- [x] `go test ./compiler3/ir/... ./compiler3/verify/... ./compiler3/frontend/... ./compiler3/emit/... -count=1`
- [x] `go test ./compiler3/build/c/... -run TestBuildSourceStrIn -count=1 -v`
- [x] `go test ./compiler3/build/... -count=1`
- [x] Smoke test on /tmp fixtures: positive match, negative match, multi-byte needle, all byte-match VM output.

### DIFF
--- a/compiler3/build/c/driver_test.go
+++ b/compiler3/build/c/driver_test.go
@@ -3041,6 +3041,45 @@ while i < len(s) {
 	}
 }
 
+// TestBuildSourceStrInBasic pins the Phase 4.2.28 `in` operator on
+// TypeStr: `needle in haystack` lowers to OpStrIn which the C target
+// emits as strstr(haystack, needle) != NULL. The fixture covers both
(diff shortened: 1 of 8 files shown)


tamnd:
tracking issue: #22049

Issue #22092, opened 2026-05-22
MEP-45 phase 3.4a: list<R> (records as list elements)


tamnd:
Tracking issue for MEP-45 Phase 3.4a: list<R>.

Phase 3.4 is the monomorphisation pass + nested-collection-types umbrella. 3.4a is its first sub-phase: `list<R>` where `R` is a user-defined record. The remaining sub-phases (3.4b nested collections, 3.4c typed-empty literal, 3.4d list / map equality) are independent and ship in follow-up issues.

## Scope

- Per-record `mochi_list_<R>` struct + 4 helpers (`_lit`, `_append`, `_index`, `_len`) emitted into the TU prologue (mirroring the scalar list helpers but TU-local, since `struct mochi_<R>` exists only inside the program).
- `ElemRecordName` parallel field on every list-bearing IR carrier (Param, Function.ReturnElemRecordName, LetStmt, VarRef, CallExpr.ResultElemRecordName, ListLit, IndexExpr, LenExpr, AppendExpr, ForEachStmt).
- Lower pass + verifier widened to accept `TypeRecord` as a list element type; nested list and list-of-map still rejected with phase-named diagnostics pointing at 3.4b.
- `IndexExpr` now propagates the receiver's record identity into the result (list<R> indexing returns a record).
- types/check_expr.go: handle `op.Field` in `checkPostfix` (was missing for `xs[0].x` and friends).
- 17 fixtures + `TestPhase3MonoListRecord` gate test.
- MEP-0045 + implementation/0045/phase-03 docs updated with sub-phase split and Decisions Made entries.

## Gate

`go test ./transpiler3/c/build/ -run TestPhase3MonoListRecord` green on host triple, plus full `./transpiler3/... ./types/...` sweep clean (no regressions).

PR: linked below.

tamnd:
PR: #22093

tamnd:
Landed. Closing.

PR #22093, opened 2026-05-22
mep-0045 phase 3.4a: list<R> (records as list elements)

tamnd:
## Summary

MEP-45 Phase 3.4a unlocks `list<R>` where `R` is a user-defined record. The phase ships:

- Per-record `mochi_list_<R>` struct + 4 static helpers (`_lit`, `_append`, `_index`, `_len`) emitted into the TU prologue, walking the program for unique record-list element types and deduping by name. Direct port of the scalar list runtime with `int64_t` replaced by `struct mochi_<R>`.
- `ElemRecordName` parallel field threaded through every list-bearing IR carrier and the verifier.
- Lower pass widened to accept `TypeRecord` as a list element type; nested-list / list-of-map still rejected with phase-named diagnostics pointing at 3.4b.
- types/check_expr.go: `op.Field` branch added to `checkPostfix` so `xs[0].x` typechecks correctly after Index / Call / Cast steps (existing gap, surfaced by 3.4a's pass-list-to-function fixtures).
- 17 fixtures + `TestPhase3MonoListRecord` gate covering literal / index / len / append / for-each / var reassign / function-call boundary (pass + return) / mixed-field records / nested for-in / conditional field access / fold across records / two record types in one program.
- MEP-0045 + impl/0045 phase-03 docs updated: Phase 3.4 split into 4 sub-phases (3.4a / 3.4b / 3.4c / 3.4d), 3.4a marked LANDED with Decisions Made section.

Tracking: #22092.

## Test plan

- [x] `go test ./transpiler3/c/build/ -run TestPhase3MonoListRecord` (17 fixtures green)
- [x] `go test ./transpiler3/... ./types/...` (full sweep clean, no regressions)
- [x] Phase 3.1 reject-test boundary still pins surface (`nested_list` now asserts "Phase 3.4")


tamnd:
Tracking issue: #22092

